PR #7, opened 2019-01-31
TSV export for record / node 

cgmeyer:
…nctions to export_record and export_node

### DIFF
--- a/gen3/submission.py
+++ b/gen3/submission.py
@@ -31,6 +31,14 @@ class Gen3Submission:
         self._auth_provider = auth_provider
         self._endpoint = endpoint
 
+    def __export_file(self, filename, output):
+        """Writes an API response to a file.
+        """
+        outfile = open(filename, "w")
+        outfile.write(output)
+        outfile.close
+        print("\nOutput written to file: "+filename)
+
     def query(self, query_txt, variables=None, max_tries=1):
         """Execute a GraphQL query against a data commons.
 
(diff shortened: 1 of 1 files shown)


gen3/submission.py
```diff
@@ -69,49 +69,67 @@ def query(self, query_txt, variables=None, max_tries=1):
...
         Examples:
             This exports a single record from the sandbox commons.
 
-            >>> Gen3Submission.export_node("DCF", "CCLE", "d70b41b9-6f90-4714-8420-e043ab8b77b9")
+            >>> Gen3Submission.export_record("DCF", "CCLE", "d70b41b9-6f90-4714-8420-e043ab8b77b9","json")
 
         """
-        api_url = "{}/api/v0/submission/{}/{}/export?ids={}&format=json".format(
-            self._endpoint, program, project, uuid
+        api_url = "{}/api/v0/submission/{}/{}/export?ids={}&format={}".format(
```
zflamig: Can we add an assert that fileformat == 'json' or 'tsv' before this since those are the only two supported types?

zflamig:
I agree with @mfshao that the function shouldn't write to a file... we can have an example of how to write out a TSV to disk separately from what the function actually does?

zflamig:
Forgot to add, once we merge this we should make a new version and bump the major version. This is a breaking change since we change the function names

gen3/submission.py
```diff
@@ -262,3 +276,11 @@ def get_graphql_schema(self):
         output = requests.get(api_url).text
         data = json.loads(output)
         return data
+
+def export_file(filename, output):
```
mfshao: should this be a private method `_export_file(self, filename, output)` rather than a plain function?

zflamig: yes please

PR #9, opened 2019-03-11
feat: add mount manifest tools

ZakirG:


gen3/mount.py
```diff
@@ -0,0 +1,164 @@
...
+        headers = {'Content-Type': 'application/json', 'Accept':'application/json', 'Authorization' : "bearer " + self.access_token }
+        request_url = self.endpoint + "manifests/file/" + urllib.parse.quote_plus(filename)
+
+        try:
+            r = requests.get(request_url, headers=headers)
+            manifest_body = r.json()['body']
+        except Exception as e:
+            raise Gen3MountError("Failed to parse manifest service response at {}\n{}\n{}".format(request_url, r.text, str(e)))
+        
+        manifests_directory = "my_manifests/"
```
mfshao: Is this path independent to each user? If so, will it causes trouble if another user tries to mount on a same computer?

PR #18, opened 2019-12-05
Feat/indexing

Avantol13:
### New Features

- Added download and verify manifest functionality for indexd (using parallel processing and multi-threading w/ asyncio library)
- Indexd API calls now retry with exponential backoff logic

### Bug Fixes

- Fix issue with indexd class erroring out when a record doesn't exist

### DIFF
--- a/gen3/tools/indexing/download_manifest.py
+++ b/gen3/tools/indexing/download_manifest.py
@@ -3,9 +3,19 @@ Module for indexing actions for downloading a manifest of
 indexed file objects (against indexd's API). Supports
 multiple processes using Python's multiprocessing library.
 
+The default manifest format created is a Comma-Separated Value file (csv)
+with rows for every record. A header row is created with field names:
+guid,authz,acl,file_size,md5,urls
+
+Fields that are lists (like acl, authz, and urls) separate the values with spaces.
+
 Attributes:
     INDEXD_RECORD_PAGE_SIZE (int): number of records to request per page
     TMP_FOLDER (str): Folder directory for placing temporary files
+        NOTE: We have to use a temporary folder b/c Python's file writing is not
+              thread-safe so we can't have all processes writing to the same file.
+              To workaround this, we have each process write to a file and concat
+              them all post-processing.
 """
 import csv
 import glob
--- a/gen3/tools/indexing/download_manifest.py
+++ b/gen3/tools/indexing/download_manifest.py
@@ -1,7 +1,7 @@
 """
 Module for indexing actions for downloading a manifest of
 indexed file objects (against indexd's API). Supports
-multiple processes using Python's multiprocessing library.
+multiple processes and coroutines using Python's asyncio library.
 
 The default manifest format created is a Comma-Separated Value file (csv)
 with rows for every record. A header row is created with field names:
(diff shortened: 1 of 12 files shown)


PlanXCyborg:
The style in this PR agrees with [`black`](https://github.com/ambv/black). :heavy_check_mark:

This formatting comment was generated automatically by a script in [uc-cdis/wool](https://github.com/uc-cdis/wool).

gen3/tools/indexing/download_manifest.py
```diff
@@ -129,15 +140,15 @@ def _write_all_index_records_to_file(commons, output_filename, num_processes):
...
     """
     Adds the given pages to the queue and starts the number of processes
     provided to consume the queue.
 
     Args:
         pages (List[int]): list of page numbers to add to the queue
         queue (multiprocessing.Queue): thread-safe multi-producer/consumer queue
-        commons (str): root domain for commons where indexd
+        commons_url (str): root domain for commons where indexd lives
         num_processes (int, optional): number of parallel python processes to use for
```
giangbui: Multiprocessing is good approach if VM/pod has multiple processors. Most of our services/jobs only run on EC2 t3.medium having only 2 processors. For that reason , maximum number of parallel jobs is only 2. Since CPUs will be in idle mode for most of the time they are waiting for http responses, it is good idea to use both multiprocessing and multi-threading? If we config 10 threads on 2-cpu VMs, each cpu will run 5 threads

gen3/tools/indexing/download_manifest.py
```diff
@@ -158,7 +169,9 @@ def _add_pages_to_queue_and_process(pages, queue, commons, num_processes):
 
     processes = []
     for x in range(num_processes):
-        p = Process(target=_get_page_and_write_records_to_file, args=(queue, commons))
+        p = Process(
+            target=_get_page_and_write_records_to_file, args=(queue, commons_url)
+        )
         p.start()
         processes.append(p)
```
giangbui: We should implement keyboard interrupt?
https://stackoverflow.com/questions/1408356/keyboard-interrupts-with-pythons-multiprocessing-pool

gen3/tools/indexing/verify_manifest.py
```diff
@@ -0,0 +1,346 @@
...
+        while row != "STOP":
+            guid = manifest_row_parsers["guid"](row)
+            authz = manifest_row_parsers["authz"](row)
+            acl = manifest_row_parsers["acl"](row)
+            file_size = manifest_row_parsers["file_size"](row)
+            md5 = manifest_row_parsers["md5"](row)
+            urls = manifest_row_parsers["urls"](row)
+
+            try:
+                actual_record = index.get_record(guid)
```
giangbui: I think we should dump all indexd records to a dictionary for lookup only. It takes for a while to build that dictionary but we can check if given record is in indexd or not very quick

gen3/tools/indexing/verify_manifest.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    Given a row from the manifest, return the field representing file's expected acls.
+
+    Args:
+        row (dict): column_name:row_value
+
+    Returns:
+        List[str]: acls for the indexd record
+    """
+    return [item for item in row.get("acl", "").split(" ") if item]
```
giangbui: return [item for item in row.get("acl", "").strip().split(" ") if item]

gen3/tools/indexing/verify_manifest.py
```diff
@@ -0,0 +1,346 @@
...
+    Given a row from the manifest, return the field representing file's expected authz
+    resources.
+
+    Args:
+        row (dict): column_name:row_value
+
+    Returns:
+        List[str]: authz resources for the indexd record
+    """
+    return [item for item in row.get("authz", "").split(" ") if item]
```
giangbui: return [item for item in row.get("authz", "").strip().split(" ") if item]

gen3/tools/indexing/verify_manifest.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    Given a row from the manifest, return the field representing file's expected urls.
+
+    Args:
+        row (dict): column_name:row_value
+
+    Returns:
+        List[str]: urls for indexd record file location(s)
+    """
+    return [item for item in row.get("urls", "").split(" ") if item]
```
giangbui: return [item for item in row.get("urls", "").strip().split(" ") if item]

gen3/tools/indexing/download_manifest.py
Avantol13: very good point, working on a threaded and multi-process version now, will update this branch

gen3/tools/indexing/verify_manifest.py
Avantol13: good point, but I would be worried about using too much memory when we have 25 million files  though and the manifest I want to verify is only say 1000. The idea with the function is that it _may_ not be all the records in indexd

gen3/tools/indexing/verify_manifest.py
Avantol13: :+1: good catch

gen3/tools/indexing/verify_manifest.py
giangbui: The indexd's limit rate now is less than 10 operations per second. If we only verify 1000, it should be fine. However we should plan for at least 100K? For 100K records, it takes about 3 hours to verify. It is too long :(. We can keep this version but I just want to make sure that we are aware of the limitation/issue.

gen3/tools/indexing/download_manifest.py
giangbui: Probably we should only implement multi-threading, the OS will distribute them to its processors

PR #19, opened 2019-12-30
Feat/indexd utils

giangbui:
Description about what this pull request does.

### New Features

- Implemented manifest indexing


manifest_indexing.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /gen3
+RUN pip install -r requirements.txt \
+    && pip install -r test-requirements.txt
+
+COPY . /gen3
+RUN cp -r src/indexclient/indexclient gen3/.
+RUN python setup.py install
+
+ENTRYPOINT [ "python" ]
+CMD [ "manifest_indexing_job.py" ]
```
Avantol13: I'm not sure this dockerfile and job belong in the SDK repo (but I could be convinced otherwise). It seems like we should have a separate location where we store this and pip install the Gen3 SDK (like an actual end-user would be installing the library) to keep this repo purely a library

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,275 @@
...
+
+UUID_FORMAT = (
+    "^.*[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+)
+MD5_FORMAT = "^[a-fA-F0-9]{32}$"
+
+
+class ThreadControl(object):
+    """
+        Class for thread synchronization
```
Avantol13: indentation issue

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,275 @@
...
+            ]
+
+        uuid = uuid.uuid4() if not fi.get("GUID") else fi.get("GUID")
+
+        doc = None
+
+        if fi.get("GUID"):
+            doc = indexclient.get(prefix + uuid)
+
+        if doc is not None:
```
Avantol13: there's an edge-case we have to consider. what if the GUID exists _and_ the manifest is requesting a different MD5 sum and/or file_size? We could 1) error out and say that you cannot change md5 sums/file_sizes without getting a new GUID and/or 2) add support for requesting a new version of a GUID. we could do this automatically here but it _would_ result in a new GUID, which could be confusing because the manifest won't have that GUID

manifest_indexing.Dockerfile
giangbui: Yeah. It is good idea to create a repos to centralize all sower jobs. All the sower jobs in the future should be configured there

gen3/tools/manifest_indexing.py
giangbui: We should think deeply for this case. GUID is not only an identifier but a reference. Usually if two GUID are different their (MD5, size) are different too. btw we have versioning too. If the file was modified, its modified version would have another GUID. It is very rare but possible case. I suggest to handle this rare case as 1). I don't think 2) is good idea since as I have mentioned GUID is very important for reference. We don't want to ask Devops to remove the newly created records if the user somehow doesn't want to maintain 2 GUIDs.

giangbui: fixed in 54c927296c47a29196341025d4b237ceae8e94f4

gen3/tools/manifest_indexing.py
giangbui: fixed in 54c927296c47a29196341025d4b237ceae8e94f4

gen3/tools/manifest_indexing.py
giangbui: fixed in e5dcd813514e7514b13ce1e33e6eb498e453e9bb

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+        logger.error("Can not read {}. Detail {}".format(manifest, e))
+        return None, None
+    
+    try:
+        headers.index("url")
+    except ValueError as e:
+        logger.error("The manifest {} has wrong format".format(manifest, e))
+        return None, None
+
+    # Generate uuid if missing
```
Avantol13: we should try to use `guid` instead of `uuid` everywhere in this file

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+        )
+    thread_control.mutexLock.release()
+
+
+def manifest_indexing(
+    manifest,
+    common_url,
+    thread_num,
+    auth=None,
+    prefix=None,
```
Avantol13: we should not support passing in a prefix. if people want one, it must be in the manifest itself in the guid column

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+Usages:
+    python manifest_indexing.py indexing --common_url https://giangb.planx-pla.net/index/ --prefix DG.1046 --manifest_path path_to_manifest --auth "admin,admin" --replace_urls False --thread_num 10
+"""
+import os
+import csv
+from functools import partial
+import logging
+from multiprocessing.dummy import Pool as ThreadPool
```
Avantol13: I know this is annoying b/c we sorta worked in parallel, but as a general comment: it would be nice if we were consistent about which libraries we're using for multiprocessing/multithreading where possible. I ended up leaning towards asyncio after Fantix's help, I think we may want to copy that same idea used for Download here for Indexing? I guess that's up for debate, but the more similar we make this tooling the easier it's going to be for new people to understand and modify (and I still have a TODO to update verify to use asyncio)

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+LOGGING_FILE = os.path.join(CURRENT_DIR, "manifest_indexing.log")
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s %(message)s",
+    handlers=[logging.FileHandler(LOGGING_FILE), logging.StreamHandler()],
+)
+
+logger = logging.getLogger()
```
Avantol13: we should avoid setting up a logger here since this will override any loggers setup outside of this. I would recommend simply importing logging and using the generate `.info` methods as usually without creating a new logger

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+    Class for thread synchronization
+    """
+
+    def __init__(self, processed_files=0, num_total_files=0):
+        self.mutexLock = threading.Lock()
+        self.num_processed_files = processed_files
+        self.num_total_files = num_total_files
+
+
+def _verify_format(s, format):
```
Avantol13: I would prefer to read this file from top to bottom, in other words, have the public functions up top and the internal functions prepended with `_` at the bottom, but that's personal preference

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+
+    if not pass_verification:
+        logger.error("The manifest is not in the correct format!!!.")
+        return None, None
+
+    return files, fieldnames
+
+
+def _write_csv(filename, files, fieldnames=None):
+
```
Avantol13: missing docstring

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,393 @@
...
+    prefix = prefix + "/" if prefix else ""
+    pool = ThreadPool(thread_num)
+
+    thread_control = ThreadControl(num_total_files=len(files))
+    part_func = partial(
+        _index_record, prefix, indexclient, replace_urls, thread_control
+    )
+
+    try:
+        pool.map_async(part_func, files).get(9999999)
```
Avantol13: what is this `9999999`?

gen3/tools/manifest_indexing.py
giangbui: Since we have limited number of connections to indexd, multithreading is the best choice for the performance. asyncio is better if we can have many connections. We have some forms of concurrency. But the question remains - when should choose which one? It really depends on the use cases. From my experience (and reading)
      CPU Bound => Multi Processing
      I/O Bound, Fast I/O, Limited Number of Connections => Multi Threading
      I/O Bound, Slow I/O, Many connections => Asyncio

gen3/tools/manifest_indexing.py
giangbui: Added in 972f9431f22112164231e6f4ccf332b1e25e34dd

gen3/tools/manifest_indexing.py
giangbui: It is async call timeout. We should have not provide it. Fixed in 43f34e300940f617609f3d47452b1febbcb5c39a

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,401 @@
+"""
+Module for indexing object files in a manifest (against indexd's API). 
+
+The default manifest format created is a Tab-Separated Value file (tsv)
+with rows for every record.
+
+Fields that are lists (like acl, authz, and urls) separate the values with commas.
```
Avantol13: if the file is tab-separated, can the default be space-delimited instead of commas for lists? that's the default for verify and download already

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,401 @@
...
+    """
+    Loop through all the files in the manifest, update/create records in indexd
+    update indexd if the url is not in the record url list or acl has changed
+
+    Args:
+        manifest(str): path to the manifest
+        common_url(str): common url
+        thread_num(int): number of threads for indexing
+        auth(Gen3Auth): Gen3 auth
+        dem(str): manifest's delimiter
```
Avantol13: missing replace_urls

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,401 @@
...
+                    )
+                )
+            else:
+                need_update = False
+                for url in urls:
+                    if not replace_urls and url not in doc.urls:
+                        doc.urls.append(url)
+                        need_update = True
+
+                if replace_urls and set(urls) != set(doc.urls):
```
Avantol13: you should do this check before looking through all the urls above. You only need to loop through if you don't do this, right? might save some time

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,401 @@
...
+        )
+    thread_control.mutexLock.release()
+
+
+def manifest_indexing(
+    manifest,
+    common_url,
+    thread_num,
+    auth=None,
+    replace_urls=False,
```
Avantol13: I think replace_urls should be the default behavior. if i am providing a manifest for indexing, I expect by default for the resulting records to be exactly what I provided

gen3/tools/manifest_indexing.py
giangbui: Good catch!

gen3/tools/manifest_indexing.py
giangbui: it supports both ta and space-delimitedb-separated

Avantol13: can we update this module docstring then, it just says commas here

gen3/tools/manifest_indexing.py
giangbui: I have just tried to use asyncio to replace multithreading but I am facing the headwind of Gen3Auth. The class was designed specifically for python requests. However, for asyncio we need to use aiohttp which provides coroutine post with auth. It will be a major change so I suggest we still keep multithreading and have TODO to update it later.

gen3/tools/manifest_indexing.py
giangbui: fixed in 513d423df92e02125f2b1e601965376aae9de572

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,429 @@
...
+    "^.*[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+)
+MD5_FORMAT = "^[a-fA-F0-9]{32}$"
+ACL_FORMAT = "^\[.*\]$"
+URL_FORMAT = "^\[.*\]$"
+
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+LOGGING_FILE = os.path.join(CURRENT_DIR, "manifest_indexing.log")
+
+logging.basicConfig(
```
Avantol13: I also don't think you should set the basic config here :sweat_smile: since this overrides any defaults the user of the library tries to set. You should just use the default logging library and not set any config

Avantol13: like just do:
```
import logging
logging.warning("some warning")
```

gen3/tools/manifest_indexing.py
```diff
@@ -0,0 +1,429 @@
...
+    "--common_url",
+    help="Root domain (url) for a commons containing indexd.",
+    required=True,
+)
+@click.option("--manifest_path", help="The path to input manifest")
+@click.option("--thread_num", type=int, help="Number of threads", default=1)
+@click.option("--api_key", help="path to api key")
+@click.option("--auth", help="basic auth")
+@click.option("--replace_urls", type=bool, help="Replace urls or not", default=False)
+def indexing(common_url, manifest_path, thread_num, api_key, auth, replace_urls):
```
Avantol13: for consistency with other functions, can we call this `index_object_manifest`?

gen3/tools/indexing/index_manifest.py
```diff
@@ -0,0 +1,488 @@
...
+                        .strip()[1:-1]
+                        .split(" ")
+                    ]
+                    if "acl" in fi and fi["acl"] != "[]"
+                    else []
+                )
+        else:
+            acl = []
+
+        # guid = uuid.uuid4() if not fi.get("GUID") else fi.get("GUID")
```
Avantol13: can we remove this?

PR #21, opened 2020-01-07
support manifest indexing

giangbui:
* update workflow

* remove wool from travis

Description about what this pull request does.

### New Features

- support indexing tsv manifest
- add dockerfile for sower job


Jenkinsfile.security
```diff
@@ -0,0 +1,6 @@
+#!groovy
```
Avantol13: why is there a jenkins file in here?

PR #29, opened 2020-03-10
Feat/asyncio verify

Avantol13:
### Improvements

- Use asyncio for verify to increase speed by using concurrency


gen3/tools/indexing/verify_manifest.py
```diff
@@ -221,113 +223,109 @@ def verify_object_manifest(
...
     Keep getting items from the queue and verifying that indexd contains the expected
     fields from that row. If there are any issues, log errors into a file. Return
     when nothing is left in the queue.
+
+    Args:
+        queue (asyncio.Queue): queue to read indexd records from
+        lock (asyncio.Semaphore): semaphones used to limit ammount of concurrent http
+            connections
+        commons_url (str): root domain for commons where indexd lives
+        output_filename (str, optional): filename for output
```
giangbui: Why you don't output to a queue and consume the queue after all the processes are done so that we don't have to open `output_filename` for writing everytime

gen3/tools/indexing/verify_manifest.py
```diff
@@ -336,77 +334,96 @@ def _verify_records_in_indexd(queue, commons_url, manifest_row_parsers):
...
-                    output = f"{guid}|file_size|expected {file_size}|actual {actual_record['size']}\n"
-                    file.write(output)
+                if file_size != actual_record["size"]:
+                    if (
+                        not file_size
+                        and file_size != 0
+                        and not actual_record["size"]
+                        and actual_record["size"] != 0
+                    ):
+                        # actual and expected are both either empty string or None
```
giangbui: I think both of them are either None or int not string so the `if` condition will never happen

tests/test_manifests.py
```diff
@@ -27,24 +27,29 @@ def test_verify_manifest(mock_index):
...
-        num_processes=3,
-        log_output_filename="test.log",
+    mock_index.return_value.async_get_record.side_effect = _async_mock_get_guid
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+
+    loop.run_until_complete(
+        async_verify_object_manifest(
+            "http://localhost",
+            manifest_file=CURRENT_DIR + "/test_manifest.csv",
```
giangbui: It would be great if `test_manifest.csv` contains a record having `acl` field with multiple spaces and we have a test for that record to make sure it was loaded correctly

gen3/tools/indexing/verify_manifest.py
Avantol13: it will if indexd blank record has None and manifest has empty string

tests/test_manifests.py
Avantol13: :+1: good idea, I will do this

PR #30, opened 2020-03-10
fix(indexing): append necessary index deployment path to commons url …

Avantol13:
…if not provided

### Bug Fixes

- append necessary indexd deployment path to commons url if not provided for indexing tool

### DIFF
--- a/gen3/tools/indexing/index_manifest.py
+++ b/gen3/tools/indexing/index_manifest.py
@@ -380,6 +380,10 @@ def index_object_manifest(
 
     """
     logging.info("Start the process ...")
+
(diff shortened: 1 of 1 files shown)


gen3/tools/indexing/index_manifest.py
```diff
@@ -381,8 +381,14 @@ def index_object_manifest(
     """
     logging.info("Start the process ...")
```
giangbui: Need to initialize service_location?

Avantol13: oh, whoops, yeah

PR #31, opened 2020-03-10
minor update to indexing docs

jiaqi216:
Pass Base URL + "/index" for `index_object_manifest` so it can make calls to the full path for indexd at`{commons_url}/index/index`


jiaqi216:
This is no longer needed with PR #30 merged - closing

PR #35, opened 2020-03-25
Feat/metadata

Avantol13:
### New Features

- Gen3Metadata class for interacting with the new metadata service
- Metadata ingestion tooling for adding to the metadata service given a file
- async query functions added to Gen3Index class

### Bug Fixes
- fix missing imports in utils.py
- make sure to raise for errors on requests responses everywhere

### Improvements
- remove unused imports and stale code
- centralize backoff config and functions to utils file
- update readme with metadata details and cleanup other examples


gen3/metadata.py
```diff
@@ -0,0 +1,224 @@
...
+        # if running locally, mds is deployed by itself without a location relative
+        # to the commons
+        if "http://localhost" in endpoint:
+            service_location = ""
+
+        if not endpoint.endswith(service_location):
+            endpoint += "/" + service_location
+
+        self.endpoint = endpoint.rstrip("/")
+        self.admin_endpoint = endpoint.rstrip("/") + "-admin"
```
vpsx, line 52: If running locally, this lands you with something like `localhost:8000-admin` haha. Maybe needs something like `if "http://localhost" in endpoint: self.admin_endpoint = self.endpoint`?

vpsx, line 52: If running locally, this lands you with something like `localhost:8000-admin` haha. Maybe needs something like `if "http://localhost" in endpoint: self.admin_endpoint = self.endpoint`?

vpsx: There is a similar issue in tools/indexing/index_manifest.py `index_object_manifest`, I think, where if running locally, the "service location" for `indexclient` is changed properly but the host is not, so it still ends up being localhost + whatever port is used by `commons_url`; you should probably be able to specify a different host/port

Avantol13: :man_facepalming: yup, I'll take a look. good catch

Avantol13: I the indexing one is fine, getting a patch for the metadata one. If localhost is provided for indexing one, it remains localhost which is what IndexClient expects. Unless I'm reading into this wrong

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+    manifest_file,
+    metadata_source,
+    auth,
+    manifest_file_delimiter,
+    max_concurrent_requests,
+    output_filename,
+    get_guid_from_file,
+):
+    """
+    Getting metadata and writing to a file. This function
```
vpsx, line 205: Doesn't this write to MDS rather, and write _logs_ to a file?

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+        for row in reader:
+            new_row = {}
+            for key, value in row.items():
+                # I know this looks crazy, DictReader is doing goofy things when
+                # column contains a JSON-like string so we're trying to fix it here
+                # Basically make sure the resulting column is something that we can
+                # later json.loads().
+                # remove redudant quoting
+                new_row[key.strip()] = (
+                    value.strip().strip("'").strip('"').replace("''", "'")
```
vpsx, line 245: :joy: I tried to reproduce these goofy things with a JSON column and I couldn't--can you post an example? :pray:

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+                )
+            await queue.put(new_row)
+
+    for _ in range(0, int(max_concurrent_requests + (max_concurrent_requests / 4))):
```
vpsx, line 249: Could you explain the reasoning behind this range upper? What's the divide by 4 thing about?

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+        await queue.put("DONE")
```
vpsx, line 250: You probably had a good reason to do it this way, but could you explain why in `_parse_from_queue()`, instead of doing ```row = await queue.get()``` and then ```while row != "DONE"```, and depend on all these "DONE"s in the queue to tell the coroutines to stop, we cannot just have in `_parse_from_queue()` for example ```while not queue.empty():  row = await queue.get(); rest of loop```? The whole queue is prepared before the `gather` is called so I... think this is OK? Hopefully this question makes sense

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+                  manifest_row_parsers["guid_for_row"] to determine the GUID
+                  (usually just a specific column in the file row like "guid")
+        metadata_source (str): the name of the source of metadata (used to namespace
+            in the metadata service) ex: dbgap
+    """
+    loop = asyncio.get_event_loop()
+
+    row = await queue.get()
+
+    while row != "DONE":
```
vpsx, line 312: In the other comment re "DONE", this is the spot I mean

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    Gets a semaphore then requests a record for the given guid
+
+    Args:
+        guid (str): indexd record globally unique id
+        commons_url (str): root domain for commons where mds lives
+        lock (asyncio.Semaphore): semaphones used to limit ammount of concurrent http
+            connections
+    """
+    index = Gen3Index(commons_url)
```
vpsx, line 431: So here if I am running everything locally, my MDS (and therefore my commons_url) might be `localhost:8000` and my indexd might be `localhost:8001`, so if I want them to play together I have to change this code. Would be nice to have this be configurable but ehh it's not too important, very nitpicky. Up to you! 

(Regarding the indexing one from my last review, the problem was the port numbers haha--the ports are part of the "host name" so even if the path is fine, it still derps. But that is even more out of scope of this PR so never mind me.)

gen3/tools/metadata/ingest_manifest.py
Avantol13: whoops yeah, this is super confusing. will clean it up

gen3/tools/metadata/ingest_manifest.py
Avantol13: hah, there were some weird edge cases that would show up sometimes where if I put the EXACT number of "Done"s to stop the processes it would sometimes have 1 process still running (e.g. inconsistency in the queue or somehow). Instead of questioning it, I just put more "Done"s then processes to make sure every process stops. I will definitely clarify this in a comment :sweat_smile: and maybe just make that an int not based on the max size provided..

gen3/tools/metadata/ingest_manifest.py
Avantol13: the Dictreader for dbgap extract would sometimes result in weird quoting like: "[''Seq_DNA_SNP_CNV'']" (note that the inside is not a double quote, it's 2 single quotes). I'll rerun the dbgap extract and try this again to see if it's still happening

gen3/tools/metadata/ingest_manifest.py
Avantol13: wait. lemme look into this more, I might be confusing this with something else

Avantol13: Ooooohhh

Avantol13: I remember now, ignore my first comment. Wow, my memory is terrible. This is because the max requests at any given time could be waiting for metadata responses all at once and there's processing done before that semaphore, so this just adds a few extra processes to get through the queue up to that point of metadata requests so it's ready right away when a lock is released. Not entirely necessary but speeds things up a tiny bit to always ensure something is waiting for that lock

gen3/tools/metadata/ingest_manifest.py
Avantol13: also sometimes looks like `"[""Seq_DNA_SNP_CNV"", ""WGS""]"` :man_shrugging:

gen3/tools/metadata/ingest_manifest.py
Avantol13: `You probably had a good reason to do it this way` You have too much faith in me :smile: I did not have a good reason and your way is much cleaner. idk why I went with this approach.. I will update it

gen3/tools/metadata/ingest_manifest.py
Avantol13: could look like `"{""status"": ""public"", ""experiments"": ""1"", ""runs"": ""1"", ""bases"": ""114518219102"", ""size_Gb"": ""20"", ""experiment_type"": ""WGS"", ""platform"": ""ILLUMINA"", ""center"": ""UM-TOPMed""}"`

gen3/tools/metadata/ingest_manifest.py
Avantol13: ohhhh, I see what you were saying now. Yeah, I did very minimal local testing, I think we'd need to rework a number of the Gen3 classes to get that to work right

gen3/metadata.py
```diff
@@ -0,0 +1,367 @@
...
+        """
+        async with aiohttp.ClientSession() as session:
+            url = self.admin_endpoint + f"/metadata/{guid}"
+            url_with_params = append_query_params(url, overwrite=overwrite, **kwargs)
+
+            # aiohttp only allows basic auth with their built in auth, so we
+            # need to manually add JWT auth header
+            headers = {"Authorization": self._auth_provider._get_auth_value()}
+
+            async with session.put(
```
vpsx, line 344: The MDS PUT method for this endpoint [does not support an "overwrite" arg](https://github.com/uc-cdis/metadata-service/blob/6747ae071e9fab47b6852eb24ba79ea01b49db4a/src/mds/maintain.py#L88) (I guess because PUT implies overwrite), so probably this should not take an "overwrite" arg? It's harmless either way--it just doesn't do anything--but it makes the logs more confusing because you see the param  there in the url

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,498 @@
...
+        output_queue (asyncio.Queue): queue for output
+        auth (Gen3Auth): Gen3 auth or tuple with basic auth name and password
+        get_guid_from_file (bool): whether or not to get the guid for metadata from file
+            NOTE: When this is True, will use the function in
+                  manifest_row_parsers["guid_for_row"] to determine the GUID
+                  (usually just a specific column in the file row like "guid")
+        metadata_source (str): the name of the source of metadata (used to namespace
+            in the metadata service) ex: dbgap
+    """
+    loop = asyncio.get_event_loop()
```
vpsx, line 315: I can't tell what this loop is for 😳 Is it used somewhere that I can't see?

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,498 @@
...
+
+    while not queue.empty():
+        row = await queue.get()
+
+        if get_guid_from_file:
+            guid = manifest_row_parsers["guid_for_row"](commons_url, row, lock)
+            is_indexed_file_object = await _is_indexed_file_object(
```
vpsx, line 322: Here indexd returns a 404 if it can't find a record, and then we backoff and retry N times, which is a bit derpy--see comment on DEFAULT_BACKOFF_SETTINGS somewhere else

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,498 @@
...
+                    f"Got conflict for {guid}. Let's update instead of create..."
+                )
+                await _update_metadata(guid, metadata, auth, commons_url, lock)
+                msg = f"Successfully updated {guid}"
+                logging.info(msg)
+                await output_queue.put(msg)
+        else:
+            msg = (
+                f"Did not add a metadata object for row because an invalid "
+                f"GUID was parsed: {guid}.\nRow: {row}"
```
vpsx, line 383: Super nitpicky: This could be a tiny bit clearer if it also said "or no record with this GUID was found in indexd"--right now it sounds like it was definitely a malformed GUID. But feel free to ignore

gen3/utils.py
```diff
@@ -0,0 +1,95 @@
...
+        "backoff: gave up call {func_call} after {tries} tries; exception: {exc}".format(
+            func_call=func_call_log, exc=sys.exc_info(), **details
+        )
+    )
+
+
+def exception_do_not_retry(error):
+    if (
+        str(getattr(error, "code", None)) == "409"
+        or str(getattr(error, "status_code", None)) == "409"
```
vpsx, line 82: 1. Should 404 and/or all 4xx errors be added to this condition? See comment somewhere else about querying indexd for a GUID to determine `is_indexed_file_object`. Maybe only retry when 5xx? Not sure

2. In the same scenario (404 from indexd) when I set a breakpoint here I actually found that the relevant attr was `status`, not `code` or `status_code` 😂 maybe this should have all three

README.md
```diff
@@ -341,3 +379,231 @@ if __name__ == "__main__":
...
+
+def main():
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+
+    auth = Gen3Auth(COMMONS, refresh_file="credentials.json")
+
+    # must provide a str to namespace the metadata from the file in a block in
+    # the metadata service
+    metadata_source = "dbgap"
```
vpsx, line 421: Random question--What is the expected outcome if I submit metadata with a GUID that already exists in MDS, but I set a different metadata_source? Is it supposed to be possible to have multiple sources submitting multiple blobs of metadata to one MDS record?
In any case I tried it and it just overwrites the old blob, so if that's expected then all good 👍

README.md
```diff
@@ -341,3 +379,231 @@ if __name__ == "__main__":
...
+
+    # (optional) override default indexd querying (NOTE: must be async)
+    async def _custom_query_for_associated_indexd_record_guid(commons_url, row, lock):
```
vpsx, line 536: This needs a fourth argument (normally `output_queue` if using default fn) or we get a TypeError, even if the custom fn doesn't end up using it. So maybe just add that to the signature in this example?

gen3/tools/metadata/ingest_manifest.py
vpsx: edit, not `DEFAULT_BACKOFF_SETTINGS`, but `exception_do_not_retry`

gen3/tools/metadata/ingest_manifest.py
vpsx: Aha this makes sense. Thanks for explaining!! I would say add a comment saying the same, but I think you removed it, which is more than fine with me :grin:

gen3/tools/metadata/ingest_manifest.py
vpsx: I must admit, I still haven't been able to come up with a manifest that triggers this... 😅 but OK!

gen3/tools/metadata/ingest_manifest.py
Avantol13: :grimacing: nope. idk why that's there.. removing

Avantol13: good catch :D

README.md
Avantol13: future state: you can update a nested blob without affecting other nested blobs (there's a PR for that Fantix was working on). Right now, it just overwrites

README.md
Avantol13: oh shoot, yeah, great catch

gen3/utils.py
Avantol13: oh, yeah, you're right. I'll update

gen3/tools/metadata/ingest_manifest.py
```diff
@@ -0,0 +1,498 @@
...
+
+    logging.info(f"done writing output to file {output_filename}")
+
+
+async def _parse_from_queue(
+    queue, lock, commons_url, output_queue, auth, get_guid_from_file, metadata_source
+):
+    """
+    Keep getting items from the queue and verifying that indexd contains the expected
+    fields from that row. If there are any issues, log errors into an error queue. Return
```
vpsx, line 299: Do you mean MDS and not indexd? Or if you mean indexd, do you mean check that it contains just the expected GUID from that row?

tests/test_metadata.py
```diff
@@ -0,0 +1,442 @@
...
+    assert response == expected_response
+
+
+@patch("gen3.metadata.requests.get")
+def test_query_full_metadata(requests_mock):
+    """
+    Test querying for guids with full data
+    """
+    metadata = Gen3Metadata("https://example.com")
+    guid = "1cfd6767-7775-4e0d-a4a7-d0fc9db02e1d"
```
vpsx, line 275: unused guid var is a bit misleading (looks like something is going to make an assertion about it)

tests/test_metadata.py
```diff
@@ -0,0 +1,442 @@
...
+@patch("gen3.metadata.requests.get")
+def test_query(requests_mock):
+    """
+    Test querying for guids
+    """
+    metadata = Gen3Metadata("https://example.com")
+    expected_response = ["1cfd6767-7775-4e0d-a4a7-d0fc9db02e1d"]
+
+    def _mock_request(url, **kwargs):
+        assert f"/metadata" in url
```
vpsx, line 253: could also, if you wanted, include `assert f"foo.bar=fizzbuzz" in url` and `assert f"data=False" in url`, and similarly `data=True` for `test_query_full_metadata`, but up to you. It might be a bit too much. I'm not sure. Testing SDKs is weird...

vpsx: but ya I am very fine with you leaving this as is

Avantol13: I'll update it, good idea. yeah.. testing SDKs is weird. I got done with all these tests and then I was like... are these even useful? they're barely testing anything. But the SDK is barely doing anything... so... :man_shrugging:

PR #36, opened 2020-04-02
Feat/hash

giangbui:
### New Features

- Introduce Gen3Hash for computing object hashes


gen3/hash.py
```diff
@@ -0,0 +1,58 @@
...
+            return None
+        with open(self.file_path, "rb") as f:
+            while True:
+                chunk = f.read(CHUNK_SIZE)
+                if not chunk:
+                    break
+                hash.update(chunk)
+            return {hash_type: hash.hexdigest()}
+
+    async def get_hash_async(self, hash_types, queue):
```
fantix, line 40: Given the task we try to complete here, this API feels a bit weird to me. What will its usage likely be? I think we could just use normal blocking API and the `multiprocessing` library to make things easier.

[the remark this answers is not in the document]
giangbui: Fixed in 79d5fa6703d811d055d985f76a9bce024c0b29fe

gen3/hash.py
```diff
@@ -0,0 +1,56 @@
...
+        with open(self.file_path, "rb") as f:
+            while True:
+                chunk = bytearray(CHUNK_SIZE)
+                num = f.readinto(chunk)
+                if not num:
+                    break
+                hash.update(chunk[:num])
+            return {hash_type: hash.hexdigest()}
+
+    def get_hashes(self, hash_types: str):
```
fantix, line 42:
```python
            return hash_type, hash.hexdigest()

    def get_hashes(self, hash_types: List[str]) -> Dict[str, str]:
        ...
        return dict(files_hash)
```

tests/test_hash.py
```diff
@@ -0,0 +1,18 @@
...
+from gen3.hash import Gen3Hash
+
+
+def _create_file_with_string(filename, s):
+    with open(filename, "w") as writer:
+        writer.write(s)
+
+
+def test_hash():
+    _create_file_with_string("./test.data", "This is a test")
```
fantix, line 13: https://docs.python.org/3/library/tempfile.html

fantix, line 13: perhaps we should also test files with size larger than the buffer size

giangbui: I know that lib but I forgot to use it for this test 🤦‍♂

PR #40, opened 2020-04-07
Feat/metadata merge

Avantol13:
### New Features
- new manifest merge feature for combining indexing and metadata manifest based on a configurable mapping


gen3/tools/merge.py
```diff
@@ -0,0 +1,298 @@
...
+    key_column_name = config.get("smaller_file_column_name")
+    value_column_names = [config.get("guid_column_name")]
+
+    column_to_matching_rows = {}
+    with open(manifest_file, "rt") as csvfile:
+        csvReader = csv.DictReader(csvfile, delimiter=delimiter)
+        for row in csvReader:
+            key = str(row[key_column_name]).strip()
+            column_to_matching_rows.setdefault(key, []).append(
+                {item: row.get(item) for item in value_column_names}
```
vpsx, line 114: Forgot to add this: So here, in the nonstandard case, where this smaller file is the metadata manifest: then here each row is a row from the metadata manifest, which does not have 'guid'; so what should `value_column_names` be here (which is to say, what should have been set for `manifests_mapping_config["guid_column_name"]`)? 
If it is "guid" then it seems all of these row.get()s will return None

gen3/tools/merge.py
```diff
@@ -0,0 +1,298 @@
...
+
+    Returns:
+        List: guids
+    """
+    # the name of column to return that exists as a value in the data_from_smaller_file
+    guid_column_name = config.get("guid_column_name", "guid")
+
+    # the key from the row to use for the exact match against the keys in
+    # data_from_smaller_file
+    row_key = config.get("row_column_name")
```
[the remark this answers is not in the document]
vpsx: eh shoot i neglected to provide you with the promised "other comment somewhere" -- here it is https://github.com/uc-cdis/gen3sdk-python/pull/40/files?file-filters%5B%5D=.html&file-filters%5B%5D=.py#r409167083

Avantol13:
> I am sorry about this huge and messy review  tldr high level: I'm not entirely convinced that this works in the case where the metadata manifest is smaller than the indexing manifest? Or I have done something wrong while testing the code   In any case, it would be nice if the user did not have to think about smaller file/bigger file, but if that is necessary, then I think some config var names could be made clearer and more documentation could be added to help the user get `manifests_mapping_config` set up correctly

yeah, I'm glad you said that. I was reading through this and I think the small performance tradeoff/notification is not worth destroying the readability with this weird smaller file stuff. Thank you for confirming my suspicions. I'll rework and let you know

gen3/tools/merge.py
```diff
@@ -0,0 +1,245 @@
...
+
+    # the key from the row to use for the exact match against the keys in
+    # data_from_indexing_manifest
+    row_key = config.get("row_column_name")
+    key_from_row = row.get(row_key).strip()
+
+    matching_guids = []
+    matching_keys = []
+    logging.info(
+        f"{len(data_from_indexing_manifest)} unmatched records remaining in indexing manifest file."
```
vpsx, line 73: This is very nice. I'm thinking, since you have the `data_from_indexing_manifest` at the end of the script anyway, in addition to printing the count you could also just print the remaining unmatched records? The thinking being, if the user has a reason to know how many unmatched records there were, they probably also want to know which ones?

gen3/tools/merge.py
```diff
@@ -0,0 +1,245 @@
...
+            "456": {"guid": "56e908b2-12df-434e-be9b-023edf66814b"}
+        }
+
+    Returns:
+        List: guids
+    """
+    # the name of column to return that exists as a value in the data_from_indexing_manifest
+    guid_column_name = config.get("guid_column_name")
+
+    # the key from the row to use for the exact match against the keys in
```
vpsx, line 65: partial match*?

gen3/tools/merge.py
```diff
@@ -0,0 +1,245 @@
...
+                [
+                    row.get(guid_column_name)
+                    for row in matching_rows
+                    if row.get(guid_column_name)
+                ]
+            )
+
+    # no need to search already matched records
+    for key in matching_keys:
+        del data_from_indexing_manifest[key]
```
vpsx, line 88: So this means that, if there were to be a GUID with a url (or whatever) that could have matched multiple metadata rows, then the order of the rows in the metadata manifest determines which metadata row "gets" the GUID. Are we assuming that this multiple match scenario will never happen? Else can we note somewhere prominent that the metadata order matters?

gen3/tools/merge.py
```diff
@@ -0,0 +1,245 @@
...
+        manifest_file (string)
+        delimiter (string): delimiter used to separate entries in the file. for a tsv,
+            this is \t
+
+    Returns:
+        column_to_matching_rows (dict): maps a key to a list and appends data from rows
+            with matching columns
+    """
+    key_column_name = config.get("indexing_manifest_column_name")
+    value_column_names = [config.get("guid_column_name")]
```
vpsx, line 108: Does this have to be a list? I can't figure out when this would ever be more than one item

gen3/tools/merge.py
```diff
@@ -0,0 +1,245 @@
...
+
+        logging.debug(f"beginning iteration over rows in {metadata_manifest}")
+        for row in reader:
+            guids = manifest_row_parsers["guids_for_manifest_row"](
+                row, data_from_indexing_manifest, config=manifests_mapping_config
+            )
+
+            if not guids:
+                # warning but write to output anyway
+                # TODO should we not write to output?
```
vpsx, line 189: I think writing to output anyway is fine--I guess if a user knows that there were metadata rows without matched GUIDS, but they don't care, then it would be very annoying if the script just gave up on them. Whereas if someone did care about unmatched records, there are warnings to indicate their presence. Perhaps you could note in the README that the script will warn about unmatched metadata records but will not fail and exit.

README.md
```diff
@@ -607,4 +609,180 @@ Setting `get_guid_from_file`  to `False` tells tool to try and get the guid usin
...
+
+The script is also fairly configurable depending on how you need to map between the two files.
+
+The ideal scenario is when you can map column to column between your _metadata manifest_ and _indexing manifest_ (e.g. what's in indexd).
+
+The non-ideal scenario is if you need something for partially matching one column to another. For example: if one of the indexed URLs will contain `submitted_sample_id` somewhere in the filename. In this case, the efficiency of the script becomes O(n2). If you can reliably parse out the section of the URL to match that could improve this. *tl;dr* Depending on your logic and number of rows in both files, this could be very very slow.
+
+> NOTE: By default this merge can match multiple GUIDs with the same metadata (depending on the configuration). This supports situations where there may exist metadata that applies to multiple files. For example: dbGaP sample metadata applied to both CRAM and CRAI genomic files.
```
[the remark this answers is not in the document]
Avantol13: good question, I think this is better solved by GUID linking (e.g. bundles) since the logical relationship is on the file objects. For now, the metadata for dbgap is at least automated, so if there were updates, both would get updated by the automation :man_shrugging: not ideal, but until we have a single GUID for the 2 files I don't think there's a better way

README.md
```diff
@@ -607,4 +609,177 @@ Setting `get_guid_from_file`  to `False` tells tool to try and get the guid usin
...
+*metadata manifest*: dbGaP extract file perhaps by using [this tool](https://github.com/uc-cdis/dbgap-extract):
+
+```
+submitted_sample_id, dbgap_subject_id, consent_short_name, body_site, ....
+```
+
+*indexing manifest* (perhaps by using the [download manifest tool](#download-manifest)):
+
+```
+guid, urls, authz, acl, md5, file_size, file_name
```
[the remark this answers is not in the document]
Avantol13: oh whoops it should be urls, thanks

gen3/tools/merge.py
Avantol13: I don't disagree, but the test file I ran for bdcat have a quarter million records, so printing out the remaining wouldn't scale since this log actually gets printed every single time an item is popped off the queue

gen3/tools/merge.py
Avantol13: if you wanted to populate the metadata service with more metadata that's already in indexd like filesize, md5, etc. which debatablely is unnecessary and a bad idea, but it was talked about as a possibility so I left this open for that sort of thing

Avantol13: this is to support cases where the guid is not available and you need to match against multiple fields like md5+file_size. I left it like this so writing your own custom function by copying this as a user leaves a pretty easy way to tack on additional fields without rewriting to much. maybe it'
s too much future-proofing :man_shrugging:

gen3/tools/merge.py
Avantol13: oh, that's an interesting point. I actually didn't consider that the GUID could match multiple metadata rows... the general case is the opposite. So I suppose the assumption is that that will never happen but I will definitely add some notes to the readme and clarify here and we can expand this functionality in the future if the need arises

Avantol13:
>  Much better imo! Yeah, even if we had been OK sacrificing readability for performance, I don't think it was only a performance issue, per the previous review--so I think it's fine to just remove the smaller file stuff.
> 
> I do wonder if we really want to load the indexing manifest into memory with a dict? My understanding is that the indexing manifest can be quite big right? (I don't remember if you said like 500k records or something?) I don't think the code actually ever does a lookup by key, just iterates through the dict items. So from what I can tell, the benefit is basically that you can `del` keys from the dict once an index record has been matched. I'm not sure if this outweighs having to keep the manifest in memory, but maybe I'm missing something. On a similar note, why OrderedDict over plan Dict?

It's sorta arbitrary to put either metadata manifest or indexing in memory because the difference in size will depend on too many external factors and how the user sets this up. It does do a dict lookup in the exact match case though (which is the ideal situation). The reason I leaned towards indexing manifest in memory by default is because it's possible to split the manifest of indexed data based on the acl/authz to narrow it down to parts of indexd you know the metadata is for (instead of using a whole dump of indexd every time). The other reason is that the metadata manifest has no defined column limit, so it's possible that each row could have hundreds of columns, which would end up making it pretty large in memory as well.

gen3/tools/merge.py
vpsx: Oof!

README.md
```diff
@@ -607,4 +609,180 @@ Setting `get_guid_from_file`  to `False` tells tool to try and get the guid usin
...
+    metadata_manifest = (
+        "/path/to/metadata_extract.tsv"
+    )
+
+    # what column to use as the final GUID for metadata (this MUST exist in the
+    # smaller file, which is expected to be the indexing file)
+    manifests_mapping_config["guid_column_name"] = "guid"
+
+    # what column from the "metadata file" to use for mapping
+    manifests_mapping_config["row_column_name"] = "submitted_sample_id"
```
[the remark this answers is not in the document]
vpsx: I still have feelings about this though :P

PR #46, opened 2020-05-23
fix(size): fix size column

giangbui:
### Bug Fixes
Fix size column


gen3/tools/indexing/index_manifest.py
```diff
@@ -174,11 +175,20 @@ def _get_and_verify_fileinfos_from_tsv_manifest(
                         logging.error("ERROR: {} is not in authz format", row[key])
                         pass_verification = False
 
+                if not pass_verification:
```
Avantol13, line 178: I think you have to move this after the new check where you set `pass_verification = False`

giangbui: The check is to early terminate if it detects the first row that does not pass validation. We want to print out all rows not pass? I have just fixed In edcf42fa9aa3733350c9269bc7ac1464455a89b6

gen3/tools/indexing/index_manifest.py
```diff
@@ -178,7 +179,15 @@ def _get_and_verify_fileinfos_from_tsv_manifest(
                     standardized_dict[standardized_key] = row[key]
 
                 elif key.lower() in SIZE:
```
Avantol13, line 181: why is this handled differently than the other keys above? in other words, can you just add this into the if/elif block above instead to be with all the other fields?

giangbui: because its value needs to be converted to `int`

Avantol13: but you can do that in the above check too. It just seemed strange that SIZE was by itself to me. I would imagine an if/elif/elif for all the fields (and you can convert to int if you have to) and then a final check for `if standardized_key`

gen3/tools/indexing/index_manifest.py
```diff
@@ -178,7 +179,15 @@ def _get_and_verify_fileinfos_from_tsv_manifest(
...
+                    fieldnames[fieldnames.index(key)] = "size"
                     standardized_dict["size"] = int(row[key]) if row[key] else None
+            if not {"urls", "md5", "size"}.issubset(set(standardized_dict.keys())):
```
Avantol13, line 184: is this possible? aren't we setting all the standardized keys above?

giangbui: yes, it is. The above code just tries to detect and convert a key to standardized one.

PR #47, opened 2020-05-27
feat(metadata): verify metadata manifest and fix issues with merging …

Avantol13:
…with BOM character

### New Features

- New verify metadata manifest functionality

### Bug Fixes

- Metadata merging handles BOM character


gen3/tools/metadata/verify_manifest.py
```diff
@@ -0,0 +1,279 @@
...
+        else:
+            # default, assume CSV
+            manifest_file_delimiter = ","
+
+    await _verify_all_metadata_records_in_file(
+        commons_url,
+        manifest_file,
+        manifest_file_delimiter,
+        max_concurrent_requests,
+        output_filename.split("/")[-1],
```
paulineribeyre, line 120: not that it matters much but why not use the full path? might be confusing to the user to feed it a path and get the output in the current dir

gen3/tools/metadata/verify_manifest.py
```diff
@@ -0,0 +1,279 @@
...
+            new_row = {}
+            for key, value in row.items():
+                new_row[key.strip()] = value.strip()
+            await queue.put(new_row)
+
+    await asyncio.gather(
+        *(
+            _parse_from_queue(queue, lock, commons_url, output_queue, metadata_source)
+            for x in range(
+                0, int(max_concurrent_requests + (max_concurrent_requests / 4))
```
paulineribeyre, line 173: sorry for the noob question, why `max_concurrent_requests + (max_concurrent_requests / 4)`?

gen3/tools/metadata/verify_manifest.py
```diff
@@ -0,0 +1,279 @@
...
+                if (
+                    isinstance(expected_value, Mapping)
+                    and not _are_matching_dicts(expected_value, actual_value)
+                ) or (actual_value != expected_value):
+                    output = f"{guid}|{metadata_source}.{expected_key}|expected {expected_value}|actual {actual_value}\n"
+                    await output_queue.put(output)
+                    logging.error(output)
+
+
+def _are_matching_dicts(dict_a, dict_b):
```
paulineribeyre, line 250: you could start with `if len(keys of a) != len(keys of b): return False` :D

gen3/tools/metadata/verify_manifest.py
```diff
@@ -0,0 +1,279 @@
...
+                try:
+                    expected_value = json.loads(str(expected_value))
+                except json.decoder.JSONDecodeError as exc:
+                    pass
+
+                # compare as dicts if necessary, otherwise just compare values
+                if (
+                    isinstance(expected_value, Mapping)
+                    and not _are_matching_dicts(expected_value, actual_value)
+                ) or (actual_value != expected_value):
```
paulineribeyre, line 244: i don't actually know if it'd make any difference at all in performance, but i think if you do this, you would avoid checking dict equality twice?
```
dicts_are_equal = _are_matching_dicts(expected_value, actual_value if isinstance(expected_value, Mapping) else actual_value != expected_value
if not dicts_are_equal:
    ....
```

gen3/tools/metadata/verify_manifest.py
Avantol13: this is a great question, and Zoe asked the same one when this was used before. probably deserves a comment. https://github.com/uc-cdis/gen3sdk-python/pull/35#discussion_r406319710

gen3/tools/metadata/verify_manifest.py
Avantol13: oh, nice! good idea

gen3/tools/metadata/verify_manifest.py
Avantol13: `actual_value if isinstance(expected_value, Mapping) else actual_value != expected_value` would result in a boolean being passed to `_are_matching_dicts` in the case where `actual_value != expected_value`. I could update the function, but this only checks dict equality if the `expected_value` is actually a dictionary

paulineribeyre: oh i forgot a parenthesis
```
dicts_are_equal = _are_matching_dicts(expected_value, actual_value)
                  if isinstance(expected_value, Mapping)
                  else actual_value != expected_value
if not dicts_are_equal:
    ....
```

PR #50, opened 2020-06-15
fix(merge): actually do what the comment said and add to output even …

Avantol13:
…if match doesn't have a guid

### Improvements
- even if the "indexing file" doesn't include GUIDs, this script can still be used to combine 2 manifests based on a common column


gen3/tools/merge.py
```diff
@@ -109,11 +116,18 @@ def _get_data_from_indexing_manifest(manifest_file, config, delimiter="\t", **kw
             with matching columns
     """
     key_column_name = config.get("indexing_manifest_column_name")
-    value_column_names = [config.get("guid_column_name")]
 
     column_to_matching_rows = {}
     with open(manifest_file, "rt", encoding="utf-8-sig") as csvfile:
         csvReader = csv.DictReader(csvfile, delimiter=delimiter)
+
+        value_column_names = [config.get("guid_column_name")]
```
paulineribeyre, line 124: could end up with `[None]`

gen3/tools/merge.py
```diff
@@ -166,17 +181,32 @@ def merge_guids_into_metadata(
...
+                    csvfile, delimiter=indexing_manifest_file_delimiter
+                )
+                headers.extend(indexing_reader.fieldnames)
+
+        # remove any duplicates but preserve order
+        unique_headers = []
+        for header in headers:
+            if header not in unique_headers:
+                unique_headers.append(header)
+        headers = unique_headers
```
paulineribeyre, line 208: could do `headers.extend(h for h in indexing_reader.fieldnames if h not in headers)`

gen3/tools/merge.py
Avantol13: good point :+1:

gen3/tools/merge.py
Avantol13: headers may already contain duplicates at this point, so I would have to reconstruct headers, I could to that with a comprehension, I just can't .extend like that

Avantol13: actually, I don't think I can do a comprehension b/c I need to compare against the list I'm constructing

paulineribeyre (approved):

PR #52, opened 2020-06-26
PXP-6220 Add bucket manifest merging

johnfrancismccann:
### New Features

- Add function `merge_bucket_manifests` in `gen3/tools/indexing/merge_manifests.py` to merge multiple input bucket manifests into a single output bucket manifest. The function reads manifest file rows into an in-memory dictionary which maps an `md5` value to a merged list consisting of space concatenated `url`, `size`, `md5`, and space concatenated `authz` values. After all of the input manifests’ file rows have been processed, the in-memory dictionary’s values are written to an output manifest.

### Benchmarking Parallel Approaches

I did attempt to parallelize this algorithm, but haven’t had much success in arriving at a solution more performant than the serial approach described above. Attempted parallel implementations `merge_manifests_by_reading_and_parsing_input_manifests_in_parallel_using_processes` and `merge_manifests_by_reading_manifests_in_parallel_using_threads` are included in [this benchmarking script](https://gist.github.com/johnfrancismccann/44b8833fb5ec1bbcdabf85925e7dd9d7). Running the benchmarking script on my laptop with 4 input manifests consisting of about a million file rows each produces the following results:
```
[<function merge_manifests_by_reading_manifests_in_parallel_using_threads at 0x105371400>] finished in 30.2631938457489 s
[<function merge_manifests_by_reading_and_parsing_input_manifests_in_parallel_using_processes at 0x10560f9d8>] finished in 29.416839122772217 s
[<function merge_manifests_serially_by_reading_whole_manifest_into_memory_at_once at 0x10561c598>] finished in 29.424784898757935 s
[<function merge_manifests_serially_by_reading_manifest_into_memory_one_row_at_a_time at 0x1056182f0>] finished in 29.002727031707764 s
```
I think one natural way to think about parallelizing this algorithm would be to have separate processes reading and parsing each input manifest, as is the case in `merge_manifests_by_reading_and_parsing_input_manifests_in_parallel_using_processes`. The problem with this approach is that data generated by the child process must be serialized and piped back to the parent process ([see Stackoverflow post about returning data from subprocesses](https://stackoverflow.com/questions/26514172/returning-large-objects-from-child-processes-in-python-multiprocessing)). I believe that an implementation involving a  `multiprocessing.Queue` would also suffer from this data serialization bottleneck ([see Stackoverflow post about queues](https://stackoverflow.com/questions/925100/python-queue-multiprocessing-queue-how-they-behave/925241#925241)).

Another possible approach would be to make use of threads, as in `merge_manifests_by_reading_manifests_in_parallel_using_threads`. While my understanding is that threads would not have the data serialization bottleneck, their usefulness would be limited to strictly I/O based operations as the GIL prevents concurrent parsing of manifest data in separate threads ([see Stackoverflow post about loading files in parallel](https://stackoverflow.com/questions/50479487/what-is-the-best-way-to-load-multiple-files-into-memory-in-parallel-using-python)).

### Specifying Manifests with Unique Files

I didn’t have time yet to include functionality to tell `merge_bucket_manifests` which input manifests are known to have unique files and thus wouldn’t need to be added to the `all_files` dictionary. One thing to note is that while having such functionality definitely could save a lot on memory usage, at least with the current implementation, the CPU-expensive section of `_merge_row` does not get executed for unique files.


gen3/tools/indexing/merge_manifests.py
```diff
@@ -0,0 +1,119 @@
...
+        rows_read(int): the number of rows read so far in file
+        total_row_count(int): the total number of rows in file
+
+    Returns:
+        None
+
+    """
+    progress = int(rows_read / total_row_count * 100)
+    pound_signs = progress // 10
+    print(f"\r[{'#'*(pound_signs)}{' '*(10-pound_signs)}] {progress}%", end="")
```
johnfrancismccann, line 119: I thought that the progress bar was a cool thing to add, but not sure if it's worth the extra code complexity and negative impact on performance. I benchmarked `merge_bucket_manifests` as taking `33.55s` with the progress bar while using the same input manifests used in  [the benchmarking script](https://gist.github.com/johnfrancismccann/44b8833fb5ec1bbcdabf85925e7dd9d7). That's about 15% slower than without the progress bar.

gen3/tools/indexing/merge_manifests.py
```diff
@@ -0,0 +1,119 @@
...
+    if hashh in all_rows:
+        size = row_to_merge[headers["size"]]
+        if size != all_rows[hashh][headers["size"]]:
+            raise csv.Error(
+                f"Could not merge file with {merge_column} equal to {hashh} while reading {input_manifest} because of size mismatch."
+            )
+
+        url = row_to_merge[headers["url"]]
+        if url not in all_rows[hashh][headers["url"]]:
+            all_rows[hashh][headers["url"]] += f" {url}"
```
johnfrancismccann, line 95: One thing to clarify is if each `url` value in the input manifest rows is limited to one url? For example, could an input manifest have a row with the `url` field set to `s3://bucket/randomvalue1 s3://bucket/randomvalue2`. If that's the case, the approach here will need to be changed to account for that. One possibility would be to store the merged urls in a set as opposed to a space-separated string. And the same thing goes for `authz` as well.

gen3/tools/indexing/merge_manifests.py
fantix: If we reduce the sampling frequency - raise `1000` to `10000` for example, would the impact become `1.5%`?

gen3/tools/indexing/merge_manifests.py
```diff
@@ -0,0 +1,119 @@
...
+            total_row_count = sum(1 for row in f)
+            f.seek(0)
+
+            reader = csv.reader(f, delimiter=delimiter)
+            headers = OrderedDict([(h, i) for i, h in enumerate(next(reader))])
+            for i, row in enumerate(reader):
+                _merge_row(all_rows, headers, row, merge_column)
+                if i % 1000 == 0 or i + 2 == total_row_count:
+                    _update_progress(i + 2, total_row_count)
+            logging.info("")
```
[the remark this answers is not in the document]
johnfrancismccann: This is to output a newline to stdout after `_update_progress` prints out the progress bar. Without it, the output would read as:
```
Iterating over manifests in ./input_manifests/one_hundred_megs directory
Reading, parsing, and merging files from ./input_manifests/one_hundred_megs/manifest3.tsv manifest
[##########] 100%Reading, parsing, and merging files from ./input_manifests/one_hundred_megs/manifest2.tsv manifest
[##########] 100%Reading, parsing, and merging files from ./input_manifests/one_hundred_megs/manifest1.tsv manifest
[##########] 100%Reading, parsing, and merging files from ./input_manifests/one_hundred_megs/manifest4.tsv manifest
[##########] 100%Writing merged manifest to merged-bucket-manifest.tsv
[##########] 100%
```   
Including the newline in `_update_progress` would cause a new progress bar to be printed out on a new line every time the progress is updated.

I think printing a newline would be better than logging an empty message here, but since we're probably going to take out the progress bar anyway, we won't need this.

gen3/tools/indexing/merge_manifests.py
```diff
@@ -0,0 +1,119 @@
...
+        1, "md5": 2, "authz": 3 })
+        row_to_merge(list(str)): the row to update all_rows with
+        merge_column(str): the header of the column that is used to merge rows
+        (e.g. "md5")
+
+    Returns:
+        None
+
+    """
+    hashh = row_to_merge[headers[merge_column]]
```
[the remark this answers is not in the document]
johnfrancismccann: I chose `hashh` because `hash` is a built-in function. This might not be an important consideration since the `hash` function isn't used here, but other possibilities could be `hash_`, `hash_code`, or `hash_value`.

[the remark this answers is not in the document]
Avantol13: ah, forgot about that, good point. yeah it's not a huge deal but I like `hash_value`, I think the "Python-preferred" way to handle collisions with builtins like this is to append the underscore like `hash_` :shrug:

gen3/tools/indexing/merge_manifests.py
johnfrancismccann: It doesn't appear so. With the frequency reduced from `1` in `1000` to `1` in `25000`, I measured the impact as being `10%`. The overhead of counting the number of lines in the manifest upfront is significant (that alone without printing the progress bar makes the script about `4%` slower)

PR #53, opened 2020-07-22
chore(hooks): add precommit hooks cfg, add security baseline

Avantol13:
### Improvements
- add precommit hooks cfg, add security baseline


.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+-   repo: git@github.com:Yelp/detect-secrets
+    rev: v0.13.1
+    hooks:
+    -   id: detect-secrets
+        args: ['--baseline', '.secrets.baseline']
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.5.0
+    hooks:
+    -   id: no-commit-to-branch
```
paulineribeyre, line 10: no trailing-whitespace and end-of-file-fixer?

Avantol13: no b/c the auto generated docs that get committed don't adhere to this, so if we enabled those checks then every commit would require adding all the modified auto generated docs

PR #55, opened 2020-08-12
WTS auth

paulineribeyre:
Depends on the `NAMESPACE` env var being available in the hatchery app - we can update hatchery to do that automatically in the future (namespace is currently available in hatchery's `POD_NAMESPACE` env var)

### New Features
- When working in a Gen3 Workspace, default to getting access tokens from the Workspace Token Service instead of requiring creds

### Deployment changes
- Depends on the hatchery app being deployed with `"env": { "NAMESPACE": <namespace> }`


docs/_build/html/_modules/gen3/tools/metadata/ingest_manifest.html
```diff
@@ -4,7 +4,7 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
frickjack, line 7: Why is the `docs/_build/` folder in git?

gen3/auth.py
```diff
@@ -27,20 +29,44 @@ class Gen3Auth(AuthBase):
...
+        self._refresh_file = refresh_file
+        self._refresh_token = refresh_token
+        self._wts_idp = idp
+        self._access_token = None
+
+        self._use_wts = False
+        self._wts_url = None
+
+        # if working in a Gen3 Workspace, we'll use the WTS
+        namespace = os.environ.get("NAMESPACE")
```
frickjack, line 43: Would be good to add a test case that sets the NAMESPACE environment variable ...

paulineribeyre: yes right now there are no unit tests for the Gen3Auth class :( i just tested locally, i'll had tests if i have time

docs/_build/html/_modules/gen3/tools/metadata/ingest_manifest.html
paulineribeyre: the docs use the _build folder 🤷  https://uc-cdis.github.io/gen3sdk-python/_build/html/index.html

paulineribeyre: they are autogenerated by a GH workflow

PR #56, opened 2020-08-19
(PXP-6498): add bucket manifest format verification

johnfrancismccann:
Jira Ticket: [PXP-6498](https://ctds-planx.atlassian.net/browse/PXP-6498)

### New Features

- Add new `is_valid_manifest_format` function to `gen3.tools.indexing` module which validates the contents of a manifest of file objects and logs all errors found. Each logged error message includes a description along with the line number and column in which the error occurred. Validation of md5, file size, url, and authz values is supported.


gen3/tools/indexing/verify_manifest_format.py
```diff
@@ -0,0 +1,410 @@
...
+            (e.g. maybe validation on a column with the name "size" is not
+            desired)
+
+            example:
+            {
+                "custom md5 name": Columns.MD5,
+                "file_size": Columns.SIZE,
+                "custom url name": Columns.URL,
+                "custom authz name": Columns.AUTHZ
+            }
```
johnfrancismccann, line 71: I think using enums that are defined in a separate class to the `Validator` classes ended up being a bit messy. For some reason, I felt that it might be advantageous for the user of `is_valid_manifest_format` to not know about or be concerned with the `Validator` classes, but we still needed a way to ultimately map manifest column names to validators, and that's where enums came in. However, in reality, the user does still need to be concerned with the `Validator` classes because that's where `ALLOWED_COLUMN_NAMES` is defined.

I was thinking that as a possible improvement, the `Columns` enum class could be removed, and the `Validator` class could be renamed to `Column` while leaving the `validate` method as is. The subclasses would then be `MD5Column`, `FileSizeColumn`, `URLColumn`, and `AuthZColumn`. This way, everything that a user could want to know about a column is in one place: identification, validation, and default allowed column names.

So `column_names_to_enums` would be replaced by `column_names_to_column_classes` and would look like this:
```
 {
    "custom md5 name": MD5Column,
    "file_size": FileSizeColumn,
     "custom url name": URLColumn,
     "custom authz name": AuthZColumn
}
```
I think this would be simpler for the user, and actually easier to implement in `is_valid_manifest_format`.

gen3/tools/indexing/verify_manifest_format.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+def _get_manifest_column_names_to_validators(
+    manifest_column_names, enums_to_validators, column_names_to_enums=None,
+):
+    """
+    Maps manifest column names to Validator subclass instances. If
+    column_names_to_enums is None, a given manifest column name is mapped to a
+    validator if the manifest column name exactly matches one of the
+    validator's ALLOWED_COLUMN_NAMES
```
johnfrancismccann, line 192: Potential improvement here would be to support case-insensitive matching.

gen3/tools/indexing/verify_manifest_format.py
```diff
@@ -0,0 +1,410 @@
...
+            determines which Validator subclass will be used for which manifest
+            column. note that when column_names_to_enums is provided by caller,
+            this means that no automatic column name matching occurs. for this
+            reason, in the example below, even though it is in SizeValidator's
+            ALLOWED_COLUMN_NAMES, "file_size" must still be explicitly mapped
+            to Columns.SIZE. this behavior is to account for the case in which
+            a user does not want validation performed on a column whose name
+            would otherwise automatically be mapped to a Validator subclass
+            (e.g. maybe validation on a column with the name "size" is not
+            desired)
```
johnfrancismccann, line 63: I think this could actually be handled by instead allowing the user to supply a `column_names_to_not_validate` list argument to `is_valid_manifest_format`

README.md
```diff
@@ -551,6 +552,128 @@ if __name__ == "__main__":
...
+```
+INFO:root:validating "tests/verify_manifest_format/manifests/manifest_with_many_types_of_errors.tsv" manifest
+INFO:root:mapped manifest column "md5" to "MD5Validator" class instance
+INFO:root:mapped manifest column "urls" to "URLValidator" class instance
+INFO:root:mapped manifest column "file_size" to "SizeValidator" class instance
+INFO:root:mapped manifest column "authz" to "AuthzValidator" class instance
+ERROR:root:line 2, "authz" value "invalid_authz" is invalid, expecting authz resource in format "/<resource>/<subresource>/.../<subresource>"
+ERROR:root:line 3, "file_size" value "invalid_int" is not an integer
+ERROR:root:line 4, "md5" value "invalid_md5" is invalid, expecting 32 hexadecimal characters
+ERROR:root:line 5, "urls" value "invalid_url" is invalid, expecting URL in format "<protocol>://<hostname>/<path>", with protocol being one of ['s3', 'gs']
```
themarcelor, line 618: @Avantol13 , have we ever discussed a model where these validation mechanisms produce an output that could be easily digestible by some other tool that could automatically fix the issues?
Maybe it could produce a JSON file with all the validation error (along with the existing log entries).

Just a thought.

themarcelor (dismissed):
+1

README.md
Avantol13: yeah, that's not a bad idea but it hasn't been considered previously. each one kind of uses it's own log format. In this case, we likely won't be the ones correcting the issue since the use case is "external group provides us a manifest, we run this check, and if there are errors, we send it back and ask for corrections" so I think it's okay for now, but we will definitely consider that in the future

README.md
```diff
@@ -22,6 +22,7 @@ The docs here contain general descriptions of the different pieces of the SDK an
     - [Verify Manifest](#verify-manifest)
     - [Indexing Manifest](#indexing-manifest)
     - [Merge Bucket Manifests](#merge-bucket-manifests)
+    - [Verify Manifest Format](#verify-manifest-format)
```
Avantol13, line 25:
```suggestion
    - [Validate Manifest Format](#validate-manifest-format)
```

README.md
```diff
@@ -551,6 +552,128 @@ if __name__ == "__main__":
     main()
 ```
 
+### Verify Manifest Format
```
Avantol13, line 555:
```suggestion
### Validate Manifest Format
```

Avantol13 (approved):

Avantol13 (approved):

PR #59, opened 2020-10-28
Fix submit_file function

paulineribeyre:
### Improvements
- "Gen3Submission.submit_file" improvements: do not fail silently in case of unexpected status code, add logs to troubleshoot failures


gen3/tools/indexing/tmp/6083.csv
```diff
@@ -0,0 +1,4 @@
+dg.TEST/f2a39f98-6ae1-48a5-8d48-825a0c52a22b,gs://test/test.txt s3://testaws/aws/test.txt,/programs/DEV/projects/test,DEV test,a1234567891234567890123456789012,123,
```
olivarity, line 1: Looks like tmp file got added to commit? Unless this was intentional

paulineribeyre: woops! thanks

olivarity (approved):

PR #60, opened 2020-11-03
(PXP-6048): Add bundles ingestion 

BinamB:
### New Features
- Added ingestion tool for bundles given a manifest.


gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+                                )
+                            )
+                            pass_verification = False
+                    record["bundles"] = bundles
+                elif key in GUID and len(value) > 0:
+                    if _verify_format(value, UUID_FORMAT):
+                        record["guid"] = value
+                    else:
+                        logging.error(
+                            "ERROR: guid: {} at row {} is in an incorrect format".format(
```
paulineribeyre, line 177: here i'd also use `key` in the log instead of "guid" 🤷‍♀️

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+                        .lstrip("[")
+                        .rstrip("]")
+                        .split(" ")
+                    )
+                    values = []
+                    for element in standard:
+                        v = element.strip().replace("'", "").replace('"', "")
+                        if len(v) > 0:
+                            values.append(v)
+                        record[key] = values
```
paulineribeyre, line 194: one too many tabs

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    replace the bundle names in the list of bundles with its associated guid
+    """
+    new_list = []
+    for bundle in bundles:
+        if _verify_format(bundle, UUID_FORMAT):
+            new_list.append(bundle)
+        else:
+            if bundle in MAIN_DICT:
+                new_list.append(MAIN_DICT[bundle])
```
paulineribeyre, line 102:
```
        elif bundle in MAIN_DICT:
            new_list.append(MAIN_DICT[bundle])
        else:
            throw error??
```

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+            )
+        else:
+            rec = resp.json()
+            guid = rec["bundle_id"]
+            logging.info("Successfully created bundle {}".format(bundle_name))
+
+            resp = drsclient.get(guid, expand=False)
+            rec = resp.json()
+
+            if resp.status_code != 200:
```
paulineribeyre, line 289: why do you need to `get` it right after `post`ing it? is it only to get the size? i'd say we should avoid making one more indexd call per bundle, if we can

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+
+    if not records:
+        return None
+
+    drsclient = DrsClient(commons_url, auth=auth, token=token)
+    total = len(records)
+    start = 0
+    # Iterate through the records list and post to indexd
+    for record in records:
+        start += 1
```
paulineribeyre, line 266: use enumerate instead of incrementing start

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+        manifest_file_delimiter(str): delimiter
+    
+    Returns:
+        record(dict): Valid bundle record to use in POST /bundle
+        records(list): List of all records to be ingested
+    NOTE: we dont do a check if indexd contains these ids since it checks that in POST /bundle
+    """
+    pass_verification = True
+    records = []
+    MAIN_DICT = {}
```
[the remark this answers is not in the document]
paulineribeyre: after reading the whole code i'm less confused about what `MAIN_DICT` is. i think it should be renamed (maybe "bundle_name_to_guid") and the docstring should describe it better

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+            record = {}
+            for key, value in row.items():
+                if key in BUNDLENAME:
+                    # make sure bundle_name exists. bundle_name is a required field.
+                    if len(value) > 0:
+                        bundle_name = value
+                        record["name"] = bundle_name
+                        MAIN_DICT[
+                            bundle_name
+                        ] = ""  # to keep track of the available bundles.
```
paulineribeyre, line 136: what happens if `bundle_name` is already in `MAIN_DICT`? how do you make sure `bundle_names` are unique in the manifest?

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+    fieldnames = [
+        "name",
+        "bundles",
+        "guid",
+        "size",
+        "checksums",
+        "description",
+        "version",
+        "aliases",
+    ]
```
paulineribeyre, line 217: depending on the format of the input, i don't think these will always match?

tests/test_import.py
```diff
@@ -29,3 +29,8 @@ def test_submission_import():
 def test_file_import():
 
     from gen3.file import Gen3File
+
+
+def test_drsclient():
+
+    import drsclient
```
paulineribeyre, line 36: do we need this

tests/bundle_tests/test_bundle_ingestion.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    records = ingest_bundle_manifest(
+        indexd_server.baseurl,
+        "./tests/bundle_tests/invalid_manifest.csv",
+        manifest_file_delimiter=",",
+        auth=("user", "user"),
+    )
+
+    assert records == None
```
paulineribeyre, line 126: should you also do `drs_client.get("dg.xxxx/590ee63d-2790-477a-bbf8-d53873ca4933")` and `drs_client.get_all(endpoint="/bundle")` to check nothing got created?

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+  {
+     "Bundle_name": "Bundle-A"
+     "Ids": [ "Bundle-B, "Bundle-C"]
+  }
+]
+"""
+
+# Pre-defined supported column names
+GUID = ["guid", "GUID"]
+BUNDLENAME = ["bundle_name"]
```
paulineribeyre, line 49: might want to accept "name" to match bundle objects :-)

gen3/tools/bundle/ingest_manifest.py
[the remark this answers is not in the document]
BinamB: i meant to rename could never come up with something better. Was supposed to be placeholder :grimacing:

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+    MAIN_DICT = {}
+    with open(manifest_file, encoding="utf-8-sig") as manifest:
+        reader = csv.DictReader(manifest, delimiter=manifest_file_delimiter)
+        row_n = 1
+        for row in reader:
+            row_n += 1
+            record = {}
+            for key, value in row.items():
+                if key in BUNDLENAME:
+                    # make sure bundle_name exists. bundle_name is a required field.
```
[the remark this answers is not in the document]
BinamB: `name` isn't a required field for bundle but for the sake of creating bundles from a manifest its required. I guess its harder to keep track of bundles that are not referenced by other bundles. I guess in the users side, who are creating the manifest, it doesnt really matter since they're naming all these bundles anyway? :shrug:

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+"""
+
+# Pre-defined supported column names
+GUID = ["guid", "GUID"]
+BUNDLENAME = ["bundle_name"]
+IDS = ["ids", "bundle_ids"]
+SIZE = ["size"]
+DESCRIPTION = ["description"]
+CHECKSUMS = ["checksums", "checksum"]
+ALIASES = ["alias", "aliases"]
```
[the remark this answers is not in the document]
BinamB: because i didnt know it existed okay!

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+
+UUID_FORMAT = (
+    r"^.*[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$"
+)
+SIZE_FORMAT = r"^[0-9]*$"
```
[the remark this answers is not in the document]
paulineribeyre: or in `manifest_columns.py`

gen3/tools/bundle/ingest_manifest.py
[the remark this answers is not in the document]
paulineribeyre: > for the sake of creating bundles from a manifest its required
but why?

> they're naming all these bundles anyway
not necessarily, since it's not a required bundle field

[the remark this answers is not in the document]
BinamB: according to the ingestion specs, in the manifest, guid is optional but bundle_name and the ids are required.
https://docs.google.com/document/d/10SrlI0OqWTdEuLdJtyx0XXDnQFL2Ryb8wxspZ3ZafOA/edit#heading=h.pyh7wcx007dw

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,320 @@
...
+                        .lstrip("[")
+                        .rstrip("]")
+                        .split(" ")
+                    )
+                    bundles = []
+                    for i in range(len(standard)):
+                        bundle = standard[i].strip().replace("'", "").replace('"', "")
+                        if bundle == record["name"]:
+                            logging.error(
+                                "Error: Bundle at row {} contains itself".format(row_n)
```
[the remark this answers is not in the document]
BinamB: Cant trust these user :suspect:

gen3/tools/bundle/ingest_manifest.py
BinamB: nah because the verification step already makes sure if the manifest is fine.

gen3/tools/bundle/ingest_manifest.py
BinamB: I was thinking for verification that its in there?

gen3/tools/bundle/ingest_manifest.py
BinamB: We write the csv using that huge list of records that we have. The keys in the records will always be the same since its the same keys we need for POST.

tests/test_import.py
BinamB: i saw a test for indexclient so just added that. I mean i guess not :shrug:

gen3/tools/indexing/manifest_columns.py
```diff
@@ -31,6 +31,16 @@
 AUTHZ_COLUMN_NAMES = ["authz"]
 AUTHZ_STANDARD_KEY = "authz"
 
+BUNDLENAME_COLUMN_NAME = ["bundle_name", "name"]
+
+IDS_COLUMN_NAME = ["ids", "bundle_ids"]
+
+DESCRIPTION_COLUMN_NAME = ["description"]
+
+CHECKSUMS_COLUMN_NAME = ["checksums", "checksum"]
```
paulineribeyre, line 40: why not use `MD5_COLUMN_NAMES`?

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+    Verify the manifest and create list of all records that needs to be ingested
+
+    Args:
+        manifest_file(str): the path to the input manifest
+        manifest_file_delimiter(str): delimiter
+    
+    Returns:
+        record(dict): Valid bundle record to use in POST /bundle
+        records(list): List of all records to be ingested
```
paulineribeyre, line 112: outdated - if the output is hard to describe, maybe use example values

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -118,67 +114,74 @@ def _create_bundle_record(manifest_file, manifest_file_delimiter="\t"):
     """
     pass_verification = True
     records = []
-    MAIN_DICT = {}
+    bundle_name_to_guid = {}  # dict containing bundle_name : bundle_guid association
     with open(manifest_file, encoding="utf-8-sig") as manifest:
         reader = csv.DictReader(manifest, delimiter=manifest_file_delimiter)
         row_n = 1
```
paulineribeyre, line 120: can remove this line

gen3/tools/bundle/ingest_manifest.py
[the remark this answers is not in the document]
paulineribeyre: ok, `bundle_name` is not required functionally (since not required in the bundle spec) nor technically (you don't need it for top level bundles) so i'm not convinced, but i'll lay off :p

gen3/tools/bundle/ingest_manifest.py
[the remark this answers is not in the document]
paulineribeyre: unresolved => `_verify_format`, `_standardize_str`

gen3/tools/bundle/ingest_manifest.py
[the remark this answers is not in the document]
BinamB: good! >.<

gen3/tools/indexing/manifest_columns.py
BinamB: DRS accepts other hashes as well.

gen3/tools/bundle/ingest_manifest.py
paulineribeyre: got it - looking at the code, i think if you use for example `checksum` instead `checksums` in the manifest, the `record` you save will use whichever you chose... i think there's a mismatch but didn't look at it in detail

gen3/tools/bundle/ingest_manifest.py
paulineribeyre: i see, yeah i'd say making fewer calls to indexd is more important 🤷‍♀️ i think we can trust indexd to have created the record when it returns a GUID

gen3/tools/indexing/manifest_columns.py
paulineribeyre: the bundle indexd documentation is not clear on the format - looking at the code i see it accepts `hashes` `checksum` or `checksums`. if using `checksum`, indexd assumes it's an md5.
the readme here doesn't describe the format of this field either.
both should be updated

gen3/tools/bundle/ingest_manifest.py
BinamB: I think you're right. It would've had `checksum` which wouldnt have been read my the drsclient.

gen3/tools/bundle/ingest_manifest.py
BinamB: We wont get the size of the bundles but i guess thats fine right?

tests/bundle_tests/test_bundle_ingestion.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+
+    records = ingest_bundle_manifest(
+        indexd_server.baseurl,
+        "./tests/bundle_tests/valid_manifest.csv",
+        manifest_file_delimiter=",",
+        auth=("user", "user"),
+    )
+
+    # 5 bundles in the manfiest
```
paulineribeyre, line 72:
```suggestion
    # 6 bundles in the manfiest
```

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -0,0 +1,293 @@
...
+        return None
+
+    drsclient = DrsClient(commons_url, auth=auth, token=token)
+    total = len(records)
+    # Iterate through the records list and post to indexd
+    for start, record in enumerate(records, 1):
+        bundle_name = record["name"]
+
+        logging.info("Posting bundle {} of {}".format(start, total))
+        record["bundles"]
```
paulineribeyre, line 263: remove this line?

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -63,25 +64,42 @@ def _replace_bundle_name_with_guid(bundles, bundle_name_to_guid):
     return new_list
 
 
+def _join_type_and_checksum(record):
+    """
+    Join checksum and their correlated type together to the following format:
+    "checksums": [{"type":"md5", "checksum":"abcdefg}, {"type":"sha256", "checksum":"abcd12345"}]
+    """
+    zipped = dict(zip(record["type"], record["checksum"]))
```
paulineribeyre, line 72: before this line:
```
if len(record["type"]) != len(record["checksum"]):
    # manifest format error
```

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -63,25 +64,42 @@ def _replace_bundle_name_with_guid(bundles, bundle_name_to_guid):
...
+    del record["type"]
+    del record["checksum"]
+    checksums = []
+    for key, value in zipped.items():
+        temp = {}
+        temp["type"] = key
```
paulineribeyre, line 78: ok i guess `dict(zip(` wasn't great advice :p
i forgot the format wasn't `{type: checksum}`
you can use this instead

```
record["checksums"] = [
    {
        "type": type,
        "checksum": checksum,
    }
    for type, checksum in zip(record["type"], record["checksum"])
]
```

gen3/tools/bundle/ingest_manifest.py
```diff
@@ -63,25 +64,42 @@ def _replace_bundle_name_with_guid(bundles, bundle_name_to_guid):
     return new_list
 
 
+def _join_type_and_checksum(record):
```
paulineribeyre, line 67: small note; it would be better practice to take `type_list` and `checksum_list` as arguments, return `checksums` and have the calling function update `record`, instead of modifying the function argument

gen3/tools/bundle/ingest_manifest.py
paulineribeyre: mm maybe this is not the right time to validate. you can also just print a warning. `zip` will work anyway, the extra values will be ignored

gen3/tools/bundle/ingest_manifest.py
BinamB: ugh you and your clean looking code :roll_eyes:

paulineribeyre (approved):

PR #63, opened 2020-12-02
Chore/poetry

Avantol13:
### New Features
- initial CLI skeleton
- CLI automatically supports pypfb and subcommands with "gen3 pfb"

### Improvements
- use poetry
- automatic PyPI deploys on master, automatic TestPyPI deploys on branches

### Dependency updates
- updated to latest versions of many deps


.travis.yml
```diff
@@ -1,20 +1,44 @@
...
+- poetry build
 deploy:
+# test pypi deployment on branches that are not master
+- provider: script
   skip_cleanup: true
-  provider: pypi
-  user: uccdistravis
-  password:
-    secure: Av0fwc8xHYWpsvamlBQH/+HsPjCW8OIpUzPN2QAOWFXY4LzJmVI7FOHfTanYCQkFjoAFu/57r0ZKf1Y4wxHDvMZtqZGJzZa8ChRgFFIVcZ8qGmXvjgjczrmR3Nxx28pAluIAtLtYSRfLKq+ckPmEmBK3r4rngErgeOnQ+DI+3vbQFOq1XgoLlLD6FQZjMAQAGnil99TK0KJblH5nXzX3lk9/+vvEclxg1Ud9K3ALWobfcuT1O2SL2MuwrDu4qw9ppNvQMr3mvRWpIFpJH1nu9AEeHq4NoM0xLrWQf+9s9WzC4bSbiVTEwsXOwZ0AZTDKWWaqEzAO5VB1QIclhI5fcOP4dDCTcGvgCmkMbV4Hdpy5oecNDoJfvBmP91quccE1pvuTFuCWXNKPscLRLSHep5zwy4DUQ2ODyIy04XYoufQOGwXVtpDhLQ7zEVWRFGcAMAjXjfU3ZBD/nolQg5cnkZN+jCos/xa+5GFsuP0ExSCSV+/ctYTi2BMXhO3kcf727DpnRYAbkj3bB56vLxjwRJXf64jMC/+mTommiRV4b3QaWxkP/mNYx58tdCCckhRRoyLnf/nOeGPP0mpyULxq2t5oU2TZZbv5H7eMWB9DiRmC+6lweMDjRh1QsMVCr9HSU0A4cSEA98VPdF7LiaKqIL68BQQ/lnLuDPjIS0fq838=
+  script: poetry publish -n -vv -r testpypi || true
```
paulineribeyre, line 23: this is cool!!

Dockerfile
```diff
@@ -2,14 +2,12 @@ from python:3.6
...
-COPY requirements.txt /gen3/requirements.txt
-COPY test-requirements.txt /gen3/test-requirements.txt
+COPY pyproject.toml /gen3/pyproject.toml
 WORKDIR /gen3
-RUN pip install -r requirements.txt \
-    && pip install -r test-requirements.txt
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
 
 COPY . /gen3
 RUN cp -r src/indexclient/indexclient gen3/.
```
paulineribeyre, line 10: i don't get this line so i checked quay and it looks like it has been broken for a while: https://quay.io/repository/cdis/gen3sdk-python?tab=builds

what is this dockerfile even for?

pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[tool.poetry.dependencies]
+python = "^3.6"
+requests = "*"
+indexclient = ">=1.6.2"
+drsclient = "<1.0.0"
+aiohttp = "*"
+backoff = "*"
+click = "*"
+pandas = "*"
```
paulineribeyre, line 28: i know they weren't pinned in the requirements.txt file, should we pin now, or do you think having the lock file is enough?

pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[tool.poetry]
+name = "gen3"
+homepage = "https://gen3.org/"
+version = "1.0.0"
```
paulineribeyre, line 4:
```suggestion
version = "2.8.0"
```

Dockerfile
Avantol13: i actually have no idea. I think to just run tests locally in a docker image? I'm happy to remove it

pyproject.toml
Avantol13: I'd like to think the lock file and unit tests are enough. if a dep breaks something then we'll notice in the PR, we can limit the versions then or update to support new version. but I could be convinced to at least make this <next_major_version assuming they all follow semantic versioning

pyproject.toml
Avantol13: oh nice catch, I'm actually gonna do a major version bump so I'll edit this to 3.0.0

.travis.yml
Avantol13: :smile:

Dockerfile
paulineribeyre: either fix it or remote it? 😆

Avantol13: I just deleted it. I don't see the point of an image in this case

Avantol13: will push the fix in a moment

paulineribeyre: i just deleted it from Quay as well

paulineribeyre (approved):

PR #64, opened 2020-12-07
fix(docs): update github action to use poetry

Avantol13:
### Improvements
- fix auto generation of docs


paulineribeyre:
`apt-get install python3-sphinx` like suggested in [sphinx docs](https://www.sphinx-doc.org/en/master/usage/installation.html)
and `sudo` like suggested in [gh action docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#administrative-privileges-of-github-hosted-runners)

paulineribeyre (approved):

PR #69, opened 2020-12-11
Fix auto sphinx docs generation

BinamB:


paulineribeyre (approved):
for future ref...
- add hardcoded path to the pythonpath so it finds the poetry installed packages
- virtualenvs.create false so that this path is predictable

PR #76, opened 2021-02-10
feat(metadata): allow specifying a metadata "type" instead of only pr…

Avantol13:
…econfigured ones

### New Features
- support specifying metadata_type when ingesting into metadata service


mpingram (approved):
LGTM

PR #83, opened 2021-04-14
PXP-7769: Tooling to simplify updating of MDS data more easily

mcannalte:
### New Features
- add discovery read, discovery publish, delete sub commands to CLI


Avantol13:
We should make sure `gen3 --help` has some help text for discovery command, it looks like it's defaulting to the same thing as "auth":

```
$ gen3 --help
Usage: gen3 [OPTIONS] COMMAND [ARGS]...

  Gen3 sdk commands

Options:
  --auth TEXT      authentication source:  "idp://wts/<idp>" is an identity
                   provider in a Gen3 workspace, "accesstoken:///<token>" is
                   an access token, otherwise a path to an api key or basename
                   of key under ~/.gen3/; default value is "credentials" if
                   ~/.gen3/credentials.json exists, otherwise
                   "idp://wts/local"

  --endpoint TEXT  commons hostname - optional if API Key given
  --help           Show this message and exit.

Commands:
  auth       Gen3 sdk auth commands
  discovery  Gen3 sdk auth commands  <------ HERE
  pfb        Commands for working with Portable Format for Biomedical Data...
  wss        Gen3 sdk workspace storage commands

```

Could have something like `Gen3 Discovery Page commands to aid in populating and modifying discovery metadata`

gen3/tools/metadata/discovery.py
```diff
@@ -0,0 +1,102 @@
+import json
+import datetime
+import tempfile
+import asyncio
+
+import pandas as pd
```
Avantol13, line 6: I know this is a big change, but can we try to avoid using pandas? It's a pretty large library and I've been trying to push for it being removed from the requirements to keep this Gen3 CLI/SDK slim. just checking my local installation, it looks like the overall deps for the sdk right now are 250MB and 47MB is pandas.

I'm open to arguments about this, but I would love to make pandas an optional install and only use it when it significantly makes things easier

gen3/tools/metadata/discovery.py
```diff
@@ -0,0 +1,102 @@
...
+    else:
+        mds = Gen3Metadata(auth_provider=auth)
+
+    metadata_df = pd.DataFrame.from_dict(
+        mds.query(
+            "_guid_type=discovery_metadata", return_full_metadata=True, limit=limit
+        ),
+        orient="index",
+    )
+    discovery_df = metadata_df.gen3_discovery.apply(pd.Series)
```
williamhaley, line 24: If I run...

```
python gen3/cli/__main__.py --auth ~/gen3credentials/credentials.dev.json discovery read
```

...against my dev env MDS with no data in it, I get an error.

```
Traceback (most recent call last):
  File "gen3/cli/__main__.py", line 55, in <module>
    main()
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/williamhaley/dev/gen3sdk-python/gen3/cli/discovery.py", line 54, in discovery_read
    auth, ctx.obj["endpoint"], limit=limit
  File "/Users/williamhaley/dev/gen3sdk-python/gen3/tools/metadata/discovery.py", line 24, in get_expanded_discovery_metadata
    discovery_df = metadata_df.gen3_discovery.apply(pd.Series)
  File "/Users/williamhaley/dev/gen3sdk-python/env/lib/python3.7/site-packages/pandas/core/generic.py", line 5179, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'gen3_discovery'
```

It works fine if the MDS has valid data though. Maybe worth adding an `if empty` sort of check somewhere? Not sure what the desired end state would be. An empty tsv, an error, nothing?

gen3/cli/discovery.py
```diff
@@ -0,0 +1,74 @@
...
+    "omit_empty",
+    is_flag=True,
+    default=False,
+    help="omit fields from empty columns if set",
+)
+@click.pass_context
+def discovery_publish(ctx, file, omit_empty):
+    """Run a discovery metadata ingestion on a given metadata TSV file with guid column"""
+    auth = ctx.obj["auth_factory"].get()
+    file = file or _metadata_file_from_auth(auth)
```
williamhaley, line 37: I wonder if it's better to be explicit about requiring a file path. I was surprised when I did a read, then a publish, that it worked automatically without me having provided an explicit path to the tsv file. It picked up what had been used from the read. I'm not sure if that's something users would expect

gen3/tools/metadata/discovery.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def get_expanded_discovery_metadata(auth, endpoint=None, limit=None) -> pd.DataFrame:
+    if endpoint:
+        mds = Gen3Metadata(auth_provider=auth, endpoint=endpoint)
+    else:
+        mds = Gen3Metadata(auth_provider=auth)
+
+    metadata_df = pd.DataFrame.from_dict(
+        mds.query(
```
williamhaley, line 19: Long term this probably won't scale. If we had an MDS with an excessive number of records, loading them all into memory (if I'm reading this right) may not work.

But I think this makes sense as-is and should work for the near future, and I may be overthinking how many records an MDS might have in reality.

gen3/cli/discovery.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+
+@click.group()
+def discovery():
+    """Gen3 sdk discovery commands"""
+    pass
+
+
+@click.command()
```
williamhaley, line 24: If it's not a huge pain, could you also add a `delete` option? I think that's something user services/other users (including myself) would want. Since discovery metadata is untyped/free-form it's common for user services and devs to upload data, realize it's not quite right, and want to try again. I think even a simple option to delete one entry at a time would be a huge help, because that could easily be scripted in bash for multiple ids.

```
mds.delete(id)
```

gen3/tools/metadata/discovery.py
mcannalte: I can't speak to how many records may be available, but it looks like this query is limited to 2000 records
https://preprod.healdata.org/mds/docs#/Query/search_metadata_metadata_get

In any case, we may be refactoring this to move away from pandas, so not keeping the full result in memory can be a priority with that change.

gen3/tools/metadata/discovery.py
mcannalte: My motivation here was to avoid the pain points around using tabular data, ex: with manifest merging.
I think removing pandas may be worthwhile, since [that design might cause difficulty with scaling down the road](https://github.com/uc-cdis/gen3sdk-python/pull/83#discussion_r615938694)

gen3/cli/discovery.py
mcannalte: This is [part of the requirements](https://docs.google.com/document/d/1XsC3t84Cn-5yP-e8cP0W69ql0a-4u2rnMfYSaMUKyjA/edit#), so I'll page @Avantol13 on this. 
My 2 cents: I did accidentally run `gen3 ... discovery publish` once or twice inadvertently

gen3/tools/metadata/discovery.py
```diff
@@ -0,0 +1,191 @@
...
+            discovery_metadata = {
+                key: _try_parse(value) for key, value in metadata_line.items()
+            }
+
+            if len(tag_columns):
+                # all columns _tag_0 -> _tag_n are pushed to a "tags" column
+                coalesced_tags = [
+                    {"name": tag_name, "category": tag_category}
+                    for tag_category, tag_name in [
+                        tag.split(": ")
```
williamhaley, line 135: I wonder if it'd be a little more forgiving to split on `:` then just `strip()` on the tokens with list comprehension to make sure we chomped any whitespace.

```
[piece.strip() for piece in tag.split(':')]
```

gen3/cli/discovery.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Delete all discovery metadata for the provided guid
+    """
+    auth = ctx.obj["auth_factory"].get()
+    try_delete_discovery_guid(auth, guid)
+
+
+discovery.add_command(discovery_read, name="read")
+discovery.add_command(discovery_publish, name="publish")
+discovery.add_command(discovery_delete, name="delete")
```
williamhaley, line 96: 👍

gen3/cli/discovery.py
```diff
@@ -0,0 +1,96 @@
...
+    auth = ctx.obj["auth_factory"].get()
+    loop = asyncio.get_event_loop()
+    if "endpoint" in ctx.obj:
+        output_file = loop.run_until_complete(
+            output_expanded_discovery_metadata(auth, ctx.obj["endpoint"], limit=limit)
+        )
+    else:
+        output_file = loop.run_until_complete(
+            output_expanded_discovery_metadata(auth, limit=limit)
+        )
```
williamhaley, line 78: I think you could collapse the conditional if you want.

```suggestion
    endpoint = ctx.obj["endpoint"] if "endpoint" in ctx.obj else None
    output_file = loop.run_until_complete(
        output_expanded_discovery_metadata(auth, endpoint, limit=limit)
    )
```

williamhaley (approved):
lgtm! Just a few minor suggestions, but they're not really necessary. Thanks for adding the delete option too!

:shipit:

PR #84, opened 2021-04-16
Feat/utils

Avantol13:
### Bug Fixes
- manifest and ingestion tools break with commas in file names, not anymore

### Improvements
- breakout utility function for getting urls from manifest string


mcannalte (approved):
LGTM!

PR #85, opened 2021-04-20
(PXP-7855): Feat/merge refactor 

Avantol13:
To reviewer: The diff is a mess b/c it's basically a rewrite. I'd recommend checking out the tests that were changed first, there are example inputs and expected outputs. The previous code failed to output the expectation.

### New Features
- ability to specify columns that should be treated like arrays (so values are combined correctly from various sources) during bucket manifest merging

### Breaking Changes
- Commas are no longer allowed as additional delimiters in a space-delimited array for indexing manifests (this was causing a bug for valid file names containing commas)

### Bug Fixes
- Fix merging edge cases where multiple rows of same hash, diff guids, weren't merging data as expected
- Fixed issue where an indexing manifest was considered invalid if it had commas in filenames in the URL column

### Improvements
- refactor merge code to be more readable and handle edge cases
- Unit tests for indexing manifest functionality better reflect edge cases around spaces & commas in URLs, duplicates, etc.


gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,310 @@ def merge_bucket_manifests(
...
-                        previous_guid_exists = True
+                updated_records = _get_updated_records(
+                    record=record,
+                    existing_records=copy.deepcopy(all_rows[record[MD5_STANDARD_KEY]]),
+                    headers=headers,
+                    continue_after_error=continue_after_error,
+                    allow_mult_guids_per_hash=allow_mult_guids_per_hash,
+                    columns_with_arrays=columns_with_arrays,
+                )
+                all_rows[record[MD5_STANDARD_KEY]] = [
```
mcannalte, line 133:
```suggestion
                all_rows[record[MD5_STANDARD_KEY]] = updated_records.values()
```

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,310 @@ def merge_bucket_manifests(
...
-                            "Found two objects with the same hash but different sizes,"
-                            f" could not merge. Details: object {record} could not be"
-                            f" merged with object {record_to_write} because {size} !="
-                            f" {record_to_write[SIZE_STANDARD_KEY]}."
-                        )
-                        logging.error(error_msg)
 
-                        if not continue_after_error:
-                            raise csv.Error(error_msg)
+def _get_updated_records(
```
mcannalte, line 157: I think it's more consistent to update `existing_records` in place, since the returned records always overwrite the previous records, and because the headers are also modified in place.

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,310 @@ def merge_bucket_manifests(
...
+                ]
 
-                    if guid:
-                        record_to_write[GUID_STANDARD_KEY] = guid
+        # for the entries where there was no GUID specified, we will add that metadata
+        # to all previous records
+        for record in records_with_no_guid:
+            updated_records = _get_updated_records(
+                record=record,
+                existing_records=copy.deepcopy(all_rows[record[MD5_STANDARD_KEY]]),
```
mcannalte, line 142: See below on updating in place vs. overwriting with a copy. If we don't update the records in place though, I think this extra copy isn't needed.

gen3/tools/indexing/merge_manifests.py
Avantol13: ooh good call, will change

gen3/tools/indexing/merge_manifests.py
Avantol13: there's a little bit of nuance here. The original is a list and what's being updated here is a dictionary based on GUID (this is so that we can merge all the information and separate on GUID more easily), consolidating things that otherwise might not. if there are multiple items in the list with the same GUID, we want those to get merged. I'm thinking through how to switch this to update in place and might push a commit, but it might be the case that this is a bit more readable, though inefficient. 

Another thing to note with efficiency here: this code only executes per hash in an edge case scenario where there are multiple GUIDs for the same hash (which _should_ never happen, but can if previous GUIDs need to be maintained alongside new ones). When this does happen, the number of duplicate hashes with diff GUIDs  would likely never exceed 3 or 4. tl;dr I'm not too worried about space inefficiency in almost all cases

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,306 @@ def merge_bucket_manifests(
...
+    """
+    updated_records = {}
+    new_guid = record.get(GUID_STANDARD_KEY)
+    new_urls = record.get(URLS_STANDARD_KEY)
+
+    # if there's no GUID, we can assume this is metadata about
+    # existing records, so update *all* of them with this information
+    if not new_guid:
+        for existing_record in existing_records:
+            existing_urls = existing_record.get(URLS_STANDARD_KEY)
```
mcannalte, line 205:
```suggestion
```
Sorry, I missed this earlier

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,306 @@ def merge_bucket_manifests(
...
-                        and guid != record_to_write.get(GUID_STANDARD_KEY)
-                    ):
-                        error_msg = (
-                            "Found two objects with the same hash but different guids,"
-                            f" could not merge. Details: object {record} could not be"
-                            f" merged with object {record_to_write} because {guid} !="
-                            f" {record_to_write.get(GUID_STANDARD_KEY)}."
-                        )
-                        logging.error(error_msg)
+            # simple case where this is the first time we've seen this hash
```
mcannalte, line 108:
```suggestion
            # simple case where this is the first time we've seen this hash
            headers.update(record_to_write.keys())
```

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,306 @@ def merge_bucket_manifests(
...
+            if record[MD5_STANDARD_KEY] not in all_rows:
+                record_to_write = copy.deepcopy(record)
+                for key in record_to_write.keys():
```
mcannalte, line 111:
```suggestion
```

gen3/tools/indexing/merge_manifests.py
```diff
@@ -70,162 +103,306 @@ def merge_bucket_manifests(
...
-                            f" {record_to_write[SIZE_STANDARD_KEY]}."
-                        )
-                        logging.error(error_msg)
 
-                        if not continue_after_error:
-                            raise csv.Error(error_msg)
+def _get_updated_records(
+    record,
+    existing_records,
+    headers,
```
mcannalte, line 156: ``
```suggestion
```

gen3/tools/indexing/merge_manifests.py
mcannalte: I think the simplicity is a valid tradeoff with such a complex feature. As a reader, it's a bit hard to follow the flow of a function that both retrieves some data and updates its arguments. 

As an alternative, I suggested a small change of removing `headers` from this function altogether and doing the headers update in just one place instead of three. The idea being that the name `_get_updated_records` seems like it wouldn't modify its args, so making this switch feels more readable. 

tldr: getter function seems like it should be parameter idempotent, could remove the headers arg and do this update closer to the other updates

mcannalte (approved):

mcannalte (approved):
LGTM!

PR #87, opened 2021-04-23
fix(add-bearer-token-header) authorization header for Bearer token for file and submission

plooploops:
### Bug Fixes
Add Authorization Header to include Bearer token:
* submit_record for submission
* get_presigned_url for file


plooploops:
Going to re-open.

PR #89, opened 2021-04-27
[DRAFT] feat(ssl): allow insecure connections

williamhaley:
### Improvements

* Allow for insecure SSL connections with --insecure. Useful in conjunction with self-signed certificates on localhost
* Start to abstract away the various HTTP clients in one place


gen3/cli/__main__.py
```diff
@@ -38,14 +39,24 @@ def get(self):
...
 @click.pass_context
-def main(ctx=None, auth_config=None, endpoint=None):
+def main(ctx=None, auth_config=None, endpoint=None, insecure=False):
     """Gen3 sdk commands"""
     ctx.ensure_object(dict)
     ctx.obj["auth_config"] = auth_config
     ctx.obj["endpoint"] = endpoint
     ctx.obj["auth_factory"] = AuthFactory(auth_config)
 
+    if insecure:
```
williamhaley, line 57: Although we _could_ put in a big "you passed in the '--insecure' flag! You should never use this in production!" I'd say we don't need that for a few reasons:
1. The libraries themselves will warn us that we're disabling verification
2. I don't think we should bend over backwards to protect users in cases like this. I think --insecure is self-explanatory especially with the --help description
3. It's an opt-in feature and off by default

gen3/cli/http_client.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+import aiohttp
+import requests
+
+verify = True
+
+
+def set_ssl_verify(value):
+    global verify
+    verify = value
```
williamhaley, line 13: Being able to set this to `False` as needed is the only way I can reliably hit localhost with our SDK. There are a number of workarounds in Python `CURL_CA_BUNDLE=""` only works for the `requests` library though, and there's really no workaround I found for `aiohttp` or `urllib3` HTTP requests.

Even if I install my cert in macOS and `Always Trust` or try to hack at the local machine certs I hit issues. I think being able to disable this is the most consistent behavior for localhost/self-signed cert testing and the simplest path for SDK users in a similar situation.

gen3/cli/http_client.py
```diff
@@ -0,0 +1,54 @@
+"""
+Contains common behaviors for HTTP interactions like disabling TLS verification
+as needed (e.g. on localhost) and to abstract away from disparate HTTP clients.
+"""
+import aiohttp
+import requests
+
+verify = True
```
paulineribeyre, line 8: use uppercase since it's a global var?

gen3/cli/http_client.py
```diff
@@ -0,0 +1,54 @@
+"""
+Contains common behaviors for HTTP interactions like disabling TLS verification
```
paulineribeyre, line 2: why is this file in the `cli` dir?

williamhaley: Just carelessness/not much consideration 😬

Probably makes sense in the root, but let me know if you think it should go somewhere else.

gen3/cli/http_client.py
williamhaley: I was trying to figure that out from pep8 🤔 https://www.python.org/dev/peps/pep-0008/#global-variable-names

> (Let's hope that these variables are meant for use inside one module only.) The conventions are about the same as those for functions.

And then functions and variable names should be lower case by default. Do we have our own internal style guidelines? It's not meant to be exported or used outside the module.

williamhaley: Maybe best if I do `_verify` to follow a "private" Python convention? https://docs.python.org/3/tutorial/classes.html#private-variables

PR #91, opened 2021-05-14
No default presigned URL protocol

paulineribeyre:
Errors that the file "does not have a location with specified protocol http" is confusing for users since they did not specify a protocol.

### Improvements
- The "Gen3File.get_presigned_url" function does not have a default "protocol" anymore. If no protocol is specified, rely on Fence and Indexd's default behavior, which is to use the first URL they find


williamhaley (approved):
:shipit:

PR #93, opened 2021-08-16
Feat/add configure

nss10:
Jira Ticket: [PXP-8581](https://ctds-planx.atlassian.net/browse/PXP-8581)

### New Features
Adding a configure command which lets the user add a profile name and attach a credentials file

### DIFF
--- a/gen3/configure.py
+++ b/gen3/configure.py
@@ -46,6 +46,6 @@ def update_config_lines(lines, profile_title, new_lines):
             configFile.writelines(lines)
     # If no profile exists then append new profile in the end
     except ValueError:
-        new_lines.insert(0, profile_title)
         with open(CONFIG_FILE_PATH, "a+") as configFile:
+            configFile.write(profile_title)
             configFile.writelines(new_lines)
--- a/gen3/configure.py
+++ b/gen3/configure.py
@@ -1,0 +1,51 @@
+import json
+import logging
+from os.path import expanduser
+from pathlib import Path
+from collections import OrderedDict
+import gen3.auth as auth_tool
+
+
+CONFIG_FILE_PATH = expanduser("~/.gen3/config")
+
+
+def get_profile_from_creds(profile, cred):
+    with open(expanduser(cred)) as f:
+        creds_from_json = json.load(f)
+        credentials = OrderedDict()
+        credentials["key_id"] = creds_from_json["key_id"]
+        credentials["api_key"] = creds_from_json["api_key"]
+        credentials["end_point"] = auth_tool.endpoint_from_token(credentials["api_key"])
+        credentials["access_key"] = auth_tool.get_access_token_with_key(credentials)
+        credentials["use_shepherd"] = ""
+        credentials["min_shepherd_version"] = ""
+    profile_line = "[" + profile + "]\n"
+    new_lines = [key + "=" + value + "\n" for key, value in credentials.items()]
+    return profile_line, new_lines
+
+
+def get_current_config_lines():
+    """ Read lines from the config file if exists, else create new config file """
+    try:
+        with open(CONFIG_FILE_PATH) as configFile:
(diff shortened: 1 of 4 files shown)


gen3/configure.py
```diff
@@ -0,0 +1,51 @@
...
+CONFIG_FILE_PATH = expanduser("~/.gen3/config")
+
+
+def get_profile_from_creds(profile, cred):
+    with open(expanduser(cred)) as f:
+        creds_from_json = json.load(f)
+        credentials = OrderedDict()
+        credentials["key_id"] = creds_from_json["key_id"]
+        credentials["api_key"] = creds_from_json["api_key"]
+        credentials["end_point"] = auth_tool.endpoint_from_token(credentials["api_key"])
```
MaribelleHGomez, line 18: Do we have documentation on what the config file looks like?

nss10: I have taken the reference from [existing implementation](https://github.com/uc-cdis/cdis-data-client/blob/cc76b65b9eece2990937d613c758ee2e3995b986/gen3-client/jwt/configure.go#L172-L178) in cdis-client to add this logic.

gen3/configure.py
```diff
@@ -0,0 +1,51 @@
...
+        logging.info(f"Config file doesn't exist at {CONFIG_FILE_PATH}, creating one")
+        return []
+
+
+def update_config_lines(lines, profile_title, new_lines):
+    """ Update config file contents with the new profile values """
+    try:
+        replace_line_index = lines.index(profile_title)
+        replace_line_index += 1
+        lines[replace_line_index : replace_line_index + len(new_lines)] = new_lines
```
paulineribeyre, line 44: i think you may be overwriting more lines than you should? for example if the previous profile config contained 3 lines and the new config contains 4 because you added `use_shepherd`

gen3/configure.py
```diff
@@ -0,0 +1,51 @@
...
+        with open(CONFIG_FILE_PATH, "w+") as configFile:
+            configFile.writelines(lines)
+    # If no profile exists then append new profile in the end
+    except ValueError:
```
paulineribeyre, line 48: could we refactor this to `if profile already exists, overwrite, else append` instead of `try to overwrite, if exception append`

gen3/configure.py
paulineribeyre: maybe it'd be easier to find the beginning and the end of the old profile, and remove it. and then append the new profile config at the end? just thinking out loud

tests/test_configure.py
```diff
@@ -0,0 +1,77 @@
...
+        if os.path.exists(test_file_name):
+            os.remove(test_file_name)
+
+    assert profile_line == expected_profile_line
+    assert lines[0] == new_lines[0]
+    assert lines[1] == new_lines[1]
+    assert lines[2] == new_lines[2]
+    assert lines[3] == new_lines[3]
+    assert lines[4] == new_lines[4]
+    assert lines[5] == new_lines[5]
```
paulineribeyre, line 63:
```suggestion
    assert lines == new_lines
```

tests/test_configure.py
```diff
@@ -0,0 +1,77 @@
...
+    file_name = str(uuid.uuid4())
+    monkeypatch.setattr(config_tool, "CONFIG_FILE_PATH", file_name)
+    try:
+        config_tool.update_config_lines(test_lines, expected_profile_line, new_lines)
+        with (open(file_name, "r")) as f:
+            assert f.readlines() == [expected_profile_line] + new_lines
+    finally:
+        if os.path.exists(file_name):
+            os.remove(file_name)
+    assert True
```
paulineribeyre, line 77:
```suggestion

```

tests/test_configure.py
paulineribeyre: makes sense, in that case you could use a loop? so that we don't have to remember update this block if we add or remove lines

PR #98, opened 2021-09-24
(PXP:8581): Add apiEndpoint and other arguments to `configure`  to avoid breaking changes

nss10:
Jira Ticket: [PXP-8581](https://ctds-planx.atlassian.net/browse/PXP-8581)

### Bug Fixes
* Replacing the existing profiles in `gen3/config` was buggy. This PR Fixes it.

### Improvements
* Added more configure arguments in order to avoid breaking changes.

### DIFF
--- a/gen3/configure.py
+++ b/gen3/configure.py
@@ -29,18 +29,22 @@ import gen3.auth as auth_tool
 CONFIG_FILE_PATH = expanduser("~/.gen3/config")
 
 
-def get_profile_from_creds(profile, cred):
+def get_profile_from_creds(
+    profile, cred, api_endpoint="", use_sheperd="", min_sheperd_version=""
+):
     with open(expanduser(cred)) as f:
         creds_from_json = json.load(f)
         credentials = OrderedDict()
         credentials["key_id"] = creds_from_json["key_id"]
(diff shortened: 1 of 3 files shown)


gen3/configure.py
```diff
@@ -60,17 +64,18 @@ def get_current_config_lines():
...
 
 def update_config_lines(lines, profile_title, new_lines):
-    """ Update config file contents with the new profile values """
+    """Update config file contents with the new profile values"""
 
     if profile_title in lines:
         profile_line_index = lines.index(profile_title)
         next_profile_index = len(lines)
-        for i in range(profile_line_index, len(lines)):
+        for i in range(profile_line_index + 1, len(lines)):
```
paulineribeyre, line 72: how come the unit tests didn't catch these 2 issues? 🤔 is there a test case we should add?

nss10: That's a good point. Let me see where did I miss it. Will get back to you on this! 👍

nss10: Ahh! It wasn't really about a missing test case but more about how the test was written. I skipped a step while arranging the test scenario. Will push the update now.

paulineribeyre (approved):

PR #101, opened 2021-10-08
Feat/logging

Avantol13:
### New Features
- verbosity flags on the CLI which map to log levels for underlying loggers

### Breaking Changes
- log format has changed (using general logging library used elsewhere in Gen3)


pyproject.toml
```diff
@@ -37,6 +37,13 @@ dosclient = { git = "https://github.com/uc-cdis/dosclient.git", tag = "1.0.0" }
 hsclient = { git = "https://github.com/uc-cdis/hsclient.git", tag = "1.0.0" }
 indexd = { git = "https://github.com/uc-cdis/indexd.git", tag = "2.14.1" }
 
+[tool.pytest.ini_options]
+minversion = "6.0"
```
paulineribeyre, line 41: why `minversion = "6.0"` here and `pytest = ">=3.2.3"` on line 33? 🤔

tests/validate_manifest_format/test_is_valid_manifest_format.py
```diff
@@ -8,13 +10,44 @@
...
+    By default, only log errors and setup a log output file to read from later
     """
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.setLevel(default_logging.ERROR)
+
+    if os.path.exists("gen3tests.logs"):
+        os.remove("gen3tests.logs")
+    logfile_handler = default_logging.FileHandler("gen3tests.logs")
+    logfile_handler.setFormatter(default_logging.Formatter(LOG_FORMAT))
+    logging.addHandler(logfile_handler)
```
paulineribeyre, line 21: cool!

pyproject.toml
Avantol13: this was to support some special caplog stuff I was testing out, I can remove it

paulineribeyre (approved):

gen3/auth.py
```diff
@@ -23,7 +26,9 @@ def decode_token(token_str):
     """
     tokenParts = token_str.split(".")
     if len(tokenParts) < 3:
-        raise Exception("invalid jwt token")
+        raise Exception(
+            f"Invalid JWT. Could not split into parts. Provided: {token_str}"
```
paulineribeyre, line 30: do we really want to print the token? 🤔

Avantol13: I thought about that... but this only happens if it's **not** a valid token (so it couldn't actually be used against the API). If it doesn't have 3 parts then it's not a valid JWT. I guess if the contents of the token are supposed to be secret and this is exposed and leaked or sent that could be bad - but the contents of tokens are not supposed to be secret since JWTs are just encoded and not encrypted.

Avantol13: I guess I was figuring that the assistance in debugging the problem with the invalid token > any (if any) risk. But I can remove this debug code (I only used it to try and figure out why the wss stuff was failing) if you think I should (I don't think we **need** it)

paulineribeyre: makes sense but it could be a typo, a real token with an extra comma at the end or something 😬 idk, people tend to copy-paste error liberally

Avantol13: that's fair, I'll take it out

paulineribeyre (dismissed):

paulineribeyre (approved):

PR #103, opened 2021-11-01
feat(profiles): simplify configuration and use standard .ini library,…

Avantol13:
… allow using --auth profile and getting api key from configuration

### New Features
- Allow passing a configure profile name in `--auth profile` to use that auth in any command

### Bug Fixes
- Don't allow multiple of the same profile name to be added to the configure file

### Improvements
- simplify configure logic to use standard .ini library in python

### DIFF
--- a/gen3/utils.py
+++ b/gen3/utils.py
@@ -2,12 +2,15 @@ import logging
 import sys
 import re
 import requests
+import configparser
+from os.path import expanduser
 
 from urllib.parse import urlunsplit
 from urllib.parse import urlencode
 from urllib.parse import urlsplit
 from urllib.parse import parse_qs
 
+CONFIG_FILE_PATH = expanduser("~/.gen3/config")
 
 UUID_FORMAT = (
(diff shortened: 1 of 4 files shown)


gen3/utils.py
```diff
@@ -2,12 +2,15 @@
...
 import requests
+import configparser
+from os.path import expanduser
 
 from urllib.parse import urlunsplit
 from urllib.parse import urlencode
 from urllib.parse import urlsplit
 from urllib.parse import parse_qs
 
+CONFIG_FILE_PATH = expanduser("~/.gen3/config")
```
mfshao, line 13: fyi `gen3-client` has also been updated recently to adpot a standard .ini library for its config files, and the config file name has been changed in there https://github.com/uc-cdis/cdis-data-client/pull/104/files#diff-bd2e7cbbf1baffd3573ff07a2de731e31dca5ef335d98bd695d8ae4baea99c5eR26

PR #107, opened 2022-01-13
PXP-9225 Submit packages

paulineribeyre:
Jira Ticket: [PXP-9225](https://ctds-planx.atlassian.net/browse/PXP-9225)

### New Features
- Additional metadata that do not fit in indexd can be submitted to the metadata service using the `gen3.tools.indexing.index_manifest.index_object_manifest` function's `submit_additional_metadata_columns` parameter, or the `gen3 objects manifest publish` CLI command's `--submit-additional-metadata-columns` flag
- Package metadata can be submitted to the metadata service during indexing, by enabling the submission of additional metadata and using `record_type = package` in the indexing manifest. The `package_contents` column can be used to describe the package


gen3/tools/indexing/index_manifest.py
```diff
@@ -454,14 +492,54 @@ def _index_record(indexclient, replace_urls, thread_control, fi):
...
 
+    # submit package metadata to metadata service
+    if index_success and fi.get(RECORD_TYPE_STANDARD_KEY) == "package":
+        try:
+            if not mds:
+                raise Exception(
+                    "Can not create package metadata when using indexd basic auth"
+                )
+
+            url = doc.urls[0]  # assume single URL for packages
```
Avantol13, line 512: :grimacing: I don't think we can do this. What if we need the package in s3 and google? I think the column still needs to be `urls`

gen3/tools/indexing/index_manifest.py
```diff
@@ -454,14 +492,54 @@ def _index_record(indexclient, replace_urls, thread_control, fi):
...
+            file_name = doc.file_name if doc.file_name else os.path.basename(url)
```
Avantol13, line 513: :shrug: idk which I actually like better. This one is debateable more readable.
```suggestion
            file_name = doc.file_name or os.path.basename(url)
```

gen3/metadata.py
```diff
@@ -397,3 +400,41 @@ def delete(self, guid, **kwargs):
         raise_for_status(response)
 
         return response.json()
+
+    def submit_package_metadata(
```
Avantol13, line 404: I'm a bit hesitant about making this specific of a function. Here's where my head's at: the concept of a package is really just metadata. The view from a pure object perspective is that this is still just a blob of data with a GUID and some metadata (it just so happens that we want to enforce some minimal structure to this metadata). But at its core, it's just more metadata. 

At the end of the day, something like `gen3 objects manifest publish` should handle a TSV with all this info + more metadata and hit the /objects API to both create an indexd record and metadata in MDS. Obviously right now that's not possible given this scope, but I want to try and set us up to easily change to that later.

So I'm wondering if we can try to generalize this a bit more or at least move it to a higher level of abstraction. Could we create a `create_object` function to replace the current `_index_record` (with the long-term intention of using it as the primary function to hit /objects)? and then for now it will just do what `_index_record` already is doing, then make a call to update metadata (where you can have a custom internal function like _get_package_metadata or something with all this custom stuff if it sees the package column). 

We should include additional columns from the manifest here to get all additional metadata (including these package columns), then update MDS with all metadata not included in standard fields in indexd:

```def get_and_verify_fileinfos_from_manifest(
    manifest_file, include_additional_columns=**True** 
```

sorry this is a novel, does it make sense though? I just want to try and **not** have an only-package-specific-metadata function, but generalize it to a function that just adds metadata (and then have downstream specific functionality to temporarily construct this information about packages until we can get /objects to behave). I think it's a subtle change, maybe I'm making way too big of a deal about this

gen3/tools/indexing/index_manifest.py
paulineribeyre: the `url` is only used to fill out the `_bucket` field in the MDS, so we need to pick one regardless 🤷‍♀️  but it will probably be less confusing if i refactor this a bit

gen3/metadata.py
```diff
@@ -397,3 +417,125 @@ def delete(self, guid, **kwargs):
         raise_for_status(response)
 
         return response.json()
+
+    def create_object(self, file_name, authz, metadata=None, aliases=None):
```
paulineribeyre, line 421: i ended up not needing this function, but i figured it'd be good to keep it in the SDK

gen3/cli/objects.py
```diff
@@ -223,8 +223,33 @@ def objects_manifest_validate_format(
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--manifest_file_delimiter",
```
Avantol13, line 227: let's do `-` delimiting words in the CLI rather than `_`

gen3/cli/objects.py
```diff
@@ -223,8 +223,33 @@ def objects_manifest_validate_format(
...
+@click.option(
+    "--out_manifest_file",
+    help="The path to output manifest (default indexing-output-manifest.csv)",
+    default="indexing-output-manifest.csv",
+)
+@click.option(
+    "--submit-additional-metadata-columns",
+    "submit_additional_metadata_columns",
+    help="If supplied, will submit additional metadata to the metadata service (default True)",
+    is_flag=True,
```
Avantol13, line 240: hmm, I'm not sure of the behavior of this in combination with `default`. You may want to make a boolean, not a flag. Because right now how would you make a call to **not** submit additional columns?

gen3/metadata.py
```diff
@@ -3,13 +3,33 @@
...
+    GUID_COLUMN_NAMES,
+    FILENAME_COLUMN_NAMES,
+    SIZE_COLUMN_NAMES,
+    MD5_COLUMN_NAMES,
+    ACLS_COLUMN_NAMES,
+    URLS_COLUMN_NAMES,
+    AUTHZ_COLUMN_NAMES,
+    PREV_GUID_COLUMN_NAMES,
+    PACKAGE_CONTENTS_STANDARD_KEY,
+    PACKAGE_CONTENTS_SCHEMA,
```
Avantol13, line 31: it might make more sense for this metadata schema to be defined in this file rather than the indexing stuff

gen3/metadata.py
```diff
@@ -397,3 +417,125 @@ def delete(self, guid, **kwargs):
...
+        self, file_name, file_size, hashes, authz, urls, contents
+    ):
+        """
+        The MDS /objects API currently expects files that have not been
+        uploaded yet. For files we only needs to index, not upload, create
+        object records manually by generating the expected object fields.
+        TODO: update the MDS objects API to not create upload URLs if the
+        relevant data is provided.
+        """
+        _url = urls[0]  # pick a URL to get the file name and bucket from
```
Avantol13, line 515: we maay want to do some minimal verification/logging if there are multiple urls and the file name is not the same. maybe a **warning**

gen3/metadata.py
```diff
@@ -397,3 +417,125 @@ def delete(self, guid, **kwargs):
...
+        if not file_name:
+            file_name = os.path.basename(_url)
+        _, file_ext = os.path.splitext(file_name)
+        parsed = urlparse(_url)
+        bucket_url = f"s3://{parsed.netloc}"
```
Avantol13, line 520: I don't think we can assume s3

gen3/cli/objects.py
paulineribeyre: True -
i think i should just not make it an option. 🤔  "gen3 objects manifest publish" and "do not submit to the MDS" doesn't make sense

Avantol13: yeah, I agree. having it as default and not alterable makes sense to me. we can always limit it in the future if we really need to

gen3/metadata.py
```diff
@@ -397,3 +437,148 @@ def delete(self, guid, **kwargs):
...
+                "file_name": file_name,
+                "created_time": now,
+                "updated_time": now,
+                "size": file_size,
+                "hashes": hashes,
+                "contents": contents or None,
+            },
+            "_resource_paths": authz,
+            "_uploader_id": uploader,
+            "_bucket": bucket_url,
```
Avantol13, line 579: I think this should be `_buckets` and be a list. even if the objects API isn't like that now, it needs to be updated

gen3/metadata.py
```diff
@@ -397,3 +437,148 @@ def delete(self, guid, **kwargs):
...
+                logging.warning(f"No URLs provided for: {submitted_metadata}")
+            for url in urls:
+                _file_name = os.path.basename(url)
+                parsed = urlparse(url)
+                _bucket_url = f"{parsed.scheme}://{parsed.netloc}"
+                if not file_name:
+                    file_name = _file_name
+                    bucket_url = _bucket_url
+                else:
+                    if file_name != _file_name or _bucket_url != bucket_url:
```
Avantol13, line 550: So it's okay if the buckets are named differently. We should keep a list of buckets in the metadata. but for file_name, I think we assume a single one, so we can do this check. e.g. I'd leave the warning for filenames but not warn on different bucket names

Avantol13 (approved):

gen3/metadata.py
nss10: For the sake of posterity for whoever reading this PR again, this corresponding function is moved to [gen3/object.py](https://github.com/uc-cdis/gen3sdk-python/blob/99b9c3e4859c21bace13a2b8475855df557f1d41/gen3/object.py#L30). 

Kudos ✌️

PR #108, opened 2022-01-19
PXP-9022: Add diff between manifests

MaribelleHGomez:
Jira Ticket: [PXP-9022](https://ctds-planx.atlassian.net/browse/PXP-9022)

- This feature is brought on by the necessity to remove indexd calls from the data replication service step for google replication

### New Features
- Takes the differential between the matching columns of two manifests of the same ext type
- Important note: rows present in prev manifest but not in succeeding manifest will not be in the differential output, those rows are treated as dep records
- Differential is output to a file


gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+import logging
+import csv
+import copy
+
+from collections import OrderedDict
+
+
+def manifest_diff(
+    directory=".",
+    files=None,
```
BinamB, line 17: We're always comparing two files right? I think something like `prev_file` `current_file` would be good.

gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+            diff_content=diff_content,
+        )
+
+
+def _precheck_manifests(
+    allow_additional_columns,
+    files=[],
+    **kwargs,
+):
+    """
```
BinamB, line 72: Comments on what this function does but you already are gonna add it. gonna write this review here just because.

gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+        List of CSVDict and headers if pass all checks
+    """
+
+    logging.info(f"Prechecking manifest files: {files}")
+
+    if not len(files) > 1:
+        logging.error("Must take difference of more than file")
+        return False
+
+    tsv_files = [file_name for file_name in files if ".tsv" in file_name.lower()]
```
BinamB, line 87: maybe not have the variable name as `tsv_files` since it could be a csv as well

gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+    if len(tsv_files) == len(files):
+        file_delimiter = "\t"
+    elif len(tsv_files) == 0:
```
BinamB, line 90: i think you should specifically check for `.csv` like you do for tsv since this would pass if all files were of the same extension like if all of them were `.img` 

I suppose, let the users know that if its not tsv or csv then we dont like it and make them convert to either one

BinamB (commented):
First round of reviews

gen3/tools/diff.py
MaribelleHGomez: so the two lines are for two input options:
if you provide a folder dir
or a string with the directory for the files with , delimiter

gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+        file_delimiter = ","
+    else:
+        logging.error("Not all files have the same extension type")
+        return False
+
+    headers = set()
+    manifest_content = []
```
Avantol13, line 119: we should use a dictionary so lookup time is O(1) and use the GUID as the key. Using a list means when we do a comparison later, we have to look over the entirety of this e.g. O(n)

gen3/tools/diff.py
```diff
@@ -0,0 +1,181 @@
...
+        Dict containing list of diff and headers
+    """
+
+    headers = manifest_content[0]["headers"]
+    diff_content = []
+    # TODO make it so that not limited to only two list being compared
+    # also make comparing logic more efficient
+    # also make it so only certain headers being compared
+    for i in manifest_content[0]["csvdict"]:
+        if i not in manifest_content[1]["csvdict"]:
```
Avantol13: this is the trouble maker here. This nested loop produces O(n^2) efficiency b/c the `in` check is over a list, so python has to iterate over everything in `manifest_content[1]["csvdict"]:`

Avantol13: if  we change these to dicts with GUIDs as keys, the lookup will be O(1)

gen3/tools/diff.py
```diff
@@ -0,0 +1,205 @@
...
+    Returns:
+        Dict containing dict of diff list and headers set
+    """
+
+    headers = manifest_content["headers"]
+    diff_content = []
+    # TODO make comparing logic more efficient
+    # ability to only have certain headers compared
+    for i in manifest_content["csvdict"][0]:
+        if i not in manifest_content["csvdict"][1]:
```
Avantol13, line 160: this is the culprit of the O(n^2) performance. The `in` is over a list so we have to iterate over everything in `manifest_content["csvdict"][1]`. If it were a dict we could lookup on GUID for O(1)

gen3/tools/diff.py
MaribelleHGomez: are we going to make this func require the manifests always have an guid "id" column

Avantol13: don't they already in practice? if not we'll have to rethink this. You can make the "unique" column configurable and if not provided, do a list like this. but if provided, it'll make a much more efficient dictionary

BinamB: `id` is always provided, I think Hannah was trying to make this code more general. But yeah +1 on making a column a key configurable.

MaribelleHGomez: I implemented a key and made it a param :)

gen3/tools/diff.py
MaribelleHGomez: updated to dict

gen3/tools/diff.py
BinamB: so the folder directory would contain both the current and the previous manifests right? I guess my point was it would be easier the read the fucntion if we had two inputs like `file_1` and `file_2` that points to the exact files/file location.

BinamB: I guess im confused on why we want to use a directory instead of pointing to two files

gen3/tools/diff.py
```diff
@@ -0,0 +1,198 @@
...
+                "id2": [{}],
+                ...
+            },
+            headers: {"header1", "header2", ...}
+        }
+    """
+
+    logging.info(f"Prechecking manifest files: {files}")
+
+    if not len(files) == 2:
```
BinamB, line 103: Kinda goes to my previous point. We dont have to check this if we just make the inputs 2 file locations

gen3/tools/diff.py
```diff
@@ -0,0 +1,198 @@
...
+
+    files.sort()
+
+    content = _precheck_manifests(
+        allow_additional_columns=allow_additional_columns,
+        files=files,
+        key_column=key_column,
+    )
+
+    if content:
```
BinamB, line 57: Can you add error handling when `content == False` and raise whatever is in the error logs in the precheck?

gen3/tools/diff.py
```diff
@@ -0,0 +1,198 @@
...
+    for manifest in files:
+        with open(manifest, "r", encoding="utf-8-sig") as csvfile:
+            csv_reader = csv.DictReader(csvfile, delimiter=file_delimiter)
+
+            field_names = csv_reader.fieldnames
+            logging.debug(f"Field names from {manifest}: {field_names}")
+            headers.update(set(field_names))
+            if not allow_additional_columns:
+                if not len(field_names) == len(headers):
+                    logging.error("Headers are not the same among manifests")
```
BinamB, line 127: Would be nice to return which headers are not the same. Since you're already making it a `set` should be easy to figure it out

BinamB (commented):
Unit tests plz

BinamB (approved):
its fine, i guess. :roll_eyes:

BinamB (dismissed):

BinamB (approved):

PR #109, opened 2022-01-20
unpacking drs packages

MichaelLukowski:
Jira Ticket: [PXP-9232](https://ctds-planx.atlassian.net/browse/PXP-9232)

### New Features
unpacking zip files that are packages in drs download

### Bug Fixes
fixed error where download from drs server where you have access token was impossible


gen3/tools/download/drs_download.py
```diff
@@ -982,6 +998,14 @@ def download(
             res = download_file_from_url(
                 url=download_url, filename=filepath, show_progress=show_progress
             )
+
+            # check if object is a zip file and a package
+            if entry.file_name.split(".")[-1] == "zip":
+                # if so expand in place
+                mds_entry = self.metadata.get(entry.object_id)
+                if mds_entry["type"] == "package":
```
paulineribeyre, line 1006:
```suggestion
                if mds_entry.get("type") == "package":
```
because non-package entries don't currently have a "type"

gen3/tools/download/drs_download.py
```diff
@@ -982,6 +998,14 @@ def download(
             res = download_file_from_url(
                 url=download_url, filename=filepath, show_progress=show_progress
             )
+
+            # check if object is a zip file and a package
+            if entry.file_name.split(".")[-1] == "zip":
```
paulineribeyre, line 1003: i'd suggest a small refactor so we can extract other formats later: check if it's a package before checking if it's a zip, instead of the opposite

maybe something like this

```
def download():
    ....
    if entry is a package:
        extract_object(filepath)

def extract_object(filepath):
    ext = ...
    if ext == zip:
        unzip(filepath)
    else:
        log a warning and don't extract
```

paulineribeyre, line 1003: `os.path.splitext` handles extensions like `.a.b`, instead of using `split`

gen3/tools/download/drs_download.py
```diff
@@ -551,6 +555,7 @@ def download_file_from_url(
         else InvisibleProgress()
     )
     try:
+        print("This is the filename", filename)
```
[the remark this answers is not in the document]
MichaelLukowski: removed

gen3/tools/download/drs_download.py
```diff
@@ -568,6 +573,13 @@ def download_file_from_url(
     return True
 
 
+def unpackage_object(filepath: str):
+    filename = str(filepath).split("/")[-1]
+    extract_dir = str(filepath).replace(filename, "")
```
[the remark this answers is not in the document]
MichaelLukowski: changed

gen3/tools/download/drs_download.py
```diff
@@ -568,6 +574,11 @@ def download_file_from_url(
     return True
 
 
+def unpackage_object(filepath: str):
```
paulineribeyre, line 577: could you add tests for this feature to `tests/download_tests` - we have a few scenarios to test, off the top of my head:
- unpack when it's a package
- don't unpack when it's not a package
- don't unpack when it's not found in MDS
- behavior when it's a package but not supported in `PACKAGE_EXTENSIONS`
- behavior when it's a package supported in `PACKAGE_EXTENSIONS` but the extraction doesn't work

tests/download_tests/test_download.py
```diff
@@ -864,3 +866,175 @@ def test_list_no_auth(
...
+                )
+                for id, item in results.items():
+                    dir_list = os.listdir(download_dir)
+                    assert "b.txt" not in dir_list and "c.txt" not in dir_list
+                    with open(download_dir.join(item.filename), "rb") as fin:
+                        assert fin.read() == base64.b64decode(
+                            download_test_files[id]["content"]
+                        )
+
+                # test everything is right but extraction doesn't work
```
paulineribeyre, line 1030: could you update this comment to explain why it's not working so we don't have to go through the test files to understand the test

docs/howto/metadataTools.md
```diff
@@ -127,110 +125,6 @@ Would result in the following metadata records in the metadata service:
 
 > NOTE: `_guid_type` is populated automatically, depending on if the provided GUID exists in indexd or not. Either `indexed_file_object` or `metadata_object`.
 
-### Index and ingest manifest
```
paulineribeyre, line 130: looks like you removed some things you shouldn't have when you rebased, the whole package ingestion commit is gone

.secrets.baseline
```diff
@@ -254,28 +120,6 @@
...
-        "is_verified": false,
-        "line_number": 6,
-        "type": "Hex High Entropy String"
-      }
-    ],
-    "tests/test_file.py": [
-      {
-        "hashed_secret": "bc5494e9e5c7bd002f295376f6130e2eced73a4a",
-        "is_verified": false,
-        "line_number": 386,
```
paulineribeyre, line 275: there's something weird going on here, this one should still be in the baseline

gen3/tools/download/drs_download.py
```diff
@@ -982,6 +997,26 @@ def download(
...
+                        try:
+                            unpackage_object(filepath)
+                        except Exception:
+                            logger.critical(
+                                f"{entry.file_name} had and issue while being unpackaged"
+                            )
+                except Exception:
+                    logger.warning(
+                        f"{entry.file_name} is not a package and will not be expanded"
+                    )
```
paulineribeyre, line 1018: can we wrap the `get()` only, and remove the warning since this is a normal use case?
also set `res` to False as mentioned [here](https://github.com/uc-cdis/gen3sdk-python/pull/109#discussion_r789837891)
```
try:
    mds_entry = self.metadata.get(entry.object_id)
except Exception:
    mds_entry = {}  # no MDS or object not in MDS

if mds_entry.get("type") == "package":
    try:
        unpackage_object(filepath)
    except Exception:
        logger.critical(
            f"{entry.file_name} had and issue while being unpackaged"
        )
        res = False
```

tests/download_tests/test_download.py
paulineribeyre: how did you make it "not work"? is the file contents bogus? that's what i mean

MichaelLukowski: I made it more clear in the test, the zip file is corrupt and can't be unzipped

paulineribeyre (dismissed):

paulineribeyre (approved):

PR #115, opened 2022-02-22
PXP-9373: Add unit tests for manifest diff script

MaribelleHGomez:
Jira Ticket: [PXP-9373](https://ctds-planx.atlassian.net/browse/PXP-9373)

### New Features
- Add unit tests for differential script
- Uses csv/tsv files in tests folder and checks for expected output defined in test file

### Dependency updates
- Change numpy to version <1.19.0 because of api incompatibility with pandas install


pyproject.toml
```diff
@@ -25,6 +25,7 @@ drsclient = "<1.0.0"
 aiohttp = "*"
 backoff = "*"
 click = "*"
+numpy = "<1.19.0"
```
Avantol13, line 28: we should remove if we don't need

MaribelleHGomez:
branch was a mess with random merges

PR #117, opened 2022-02-23
[WIP] Add object api for sdk

nss10:

### DIFF
--- a/gen3/object.py
+++ b/gen3/object.py
@@ -1,0 +1,81 @@
+from gen3 import file, metadata, auth as auth_tool, index as indexd
+
+
+class Gen3ObjectError(Exception):
+    pass
+
+
+class Gen3Object:
+    """For interacting with Gen3 object level features.
+
+    A class for interacting with the Gen3 object services.
+    Currently allows deleting of an object from the Gen3 System.
+
+    Args:
+        auth_provider (Gen3Auth): A Gen3Auth class instance.
+
+    Examples:
+        This generates the Gen3Object class pointed at the sandbox commons while
+        using the credentials.json downloaded from the commons profile page.
+
+        >>> auth = Gen3Auth(refresh_file="credentials.json")
+        ... file = Gen3File(auth)
+
+    """
+
+    def __init__(self, endpoint=None, auth_provider=None):
+        # auth_provider legacy interface required endpoint as 1st arg
+        self._auth_provider = auth_provider or endpoint
+        self._endpoint = self._auth_provider.endpoint
+
+    def delete_object(self, guid, delete_record=False):
+        """
+        Delete the all the records of the guid from all storage locations, indexd and metadata service
+
+        Args:
+            guid (str): provide a UUID for file id to delete
+        Returns:
+            text: requests.delete text result
+        """
+        fence = file.Gen3File(auth_provider=self._auth_provider)
+        meta = metadata.Gen3Metadata(auth_provider=self._auth_provider)
+        metadata_response_object = None
+        metadata_response_object = meta.query(
+            query=f"guid={guid}&data=True", return_full_metadata=True
+        )
+
+        if metadata_response_object:
+            try:
+                metadata_response_object = meta.delete(guid)
+                print(metadata_response_object)
+            except Exception as exp:
+                return f"Error in deleting object with {guid} from Metadata Service. Response -- {exp}"
+
+        index = indexd.Gen3Index(auth_provider=self._auth_provider)
+        idx_record = index.get(guid)
+
+        if not idx_record:
+            return f"No record found with guid- {guid}"
+
+        if delete_record:
+            response = fence.delete_file_locations(guid)
+            if response.status_code != 204:
+                if metadata_response_object:
+                    meta.create(guid, metadata_response_object)
+                return f"Error in deleting object with {guid} from Fence. Response -- {response}"
+        else:
+            response = index.delete_record(guid)
+            print(type(response))
+            print(response)
+        return "Success"
+
+
+if __name__ == "__main__":
+    auth = auth_tool.Gen3Auth(
+        refresh_file="/Users/saishanmukhanarumanchi/Documents/git_projects/playground/sai_creds.json"
+    )
+    gen3Obj = Gen3Object(auth_provider=auth)
+    fence = file.Gen3File(auth)
+    guid = fence.upload_file("query.py", protocol="s3")["guid"]
+    print(guid)
+    print(gen3Obj.delete_object(guid, delete_record=True))
--- a/gen3/file.py
+++ b/gen3/file.py
@@ -56,6 +56,21 @@ class Gen3File:
 
     def delete_file(self, guid):
         """
+        This method is DEPRECATED. Use delete_file_locations() instead.
+        Delete all locations of a stored data file and remove its record from indexd
+
+        Args:
+            guid (str): provide a UUID for file id to delete
+        Returns:
+            requests.Response : requests.delete result
+        """
+        api_url = "{}/user/data/{}".format(self._endpoint, guid)
+        output = requests.delete(api_url, auth=self._auth_provider).text
+
+        return output
+
+    def delete_file_locations(self, guid):
+        """
         Delete all locations of a stored data file and remove its record from indexd
 
         Args:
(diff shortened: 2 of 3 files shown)


gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+
+        >>> auth = Gen3Auth(refresh_file="credentials.json")
+        ... file = Gen3File(auth)
+
+    """
+
+    def __init__(self, endpoint=None, auth_provider=None):
+        # auth_provider legacy interface required endpoint as 1st arg
+        self._auth_provider = auth_provider or endpoint
+        self._endpoint = self._auth_provider.endpoint
```
paulineribeyre, line 30: for new modules, we shouldn't need this backwards compatibility stuff. you can take `auth_provider` as the only parameter

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def delete_object(self, guid, delete_record=False):
+        """
+        Delete the all the records of the guid from all storage locations, indexd and metadata service
```
paulineribeyre, line 34:
```suggestion
        Delete the object from indexd, metadata service and optionally all storage locations
```

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+
+        Args:
+            guid (str): provide a UUID for file id to delete
+        Returns:
+            text: requests.delete text result
```
paulineribeyre, line 39: these need an update

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        fence = file.Gen3File(auth_provider=self._auth_provider)
+        meta = metadata.Gen3Metadata(auth_provider=self._auth_provider)
+        metadata_response_object = None
```
paulineribeyre, line 43: you don't need this line

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+                metadata_response_object = meta.delete(guid)
+                logging.info(metadata_response_object)
+            except Exception as exp:
+                return f"Error in deleting object with {guid} from Metadata Service. Exception -- {exp}"
+
+        index = indexd.Gen3Index(auth_provider=self._auth_provider)
+        idx_record = index.get(guid)
+
+        if not idx_record:
+            return f"No record found with guid- {guid}"
```
paulineribeyre, line 57: i don't think you need to GET the indexd record 🤔  if it doesn't exist, something is wrong (you shouldn't be calling the object deletion function for a non-object record), the deletion will fail and the metadata record should be re-created

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if delete_record:
+            response = fence.delete_file_locations(guid)
+            if response.status_code != 204:
+                if metadata_response_object:
+                    meta.create(guid, metadata_response_object)
+                return f"Error in deleting object with {guid} from Fence. Response -- {response}"
```
paulineribeyre, line 64: re-creating the MDS record should also happen if the indexd delete fails, not only if the fence delete fails. See the pseudo code https://cdis.slack.com/archives/C033FPMPSKE/p1645565587258179

gen3/object.py
```diff
@@ -0,0 +1,79 @@
...
+        meta = metadata.Gen3Metadata(auth_provider=self._auth_provider)
+        metadata_response_object = None
+        metadata_response_object = meta.query(f"guid={guid}&data=True", True)
+
+        if metadata_response_object:
+            try:
+                metadata_response_object = meta.delete(guid)
+                logging.info(metadata_response_object)
+            except Exception as exp:
+                return f"Error in deleting object with {guid} from Metadata Service. Exception -- {exp}"
```
paulineribeyre, line 51: the function should raise exceptions in case of errors instead of returning strings. similarly, success messages should be logged, not returned

gen3/object.py
nss10: Do we still need the parameter `self._endpoint` variable in the class?

gen3/object.py
nss10: Ohh! Wait.. I see it now, [this line](https://github.com/uc-cdis/gen3sdk-python/blob/6d12af7b0b7c00b7bbdb746739a6008e1f97c2eb/gen3/index.py#L729) throws an exception after a few retires. Got it, will update it.

gen3/object.py
paulineribeyre: then like i said the deletion will fail with a 404
i would rather we avoid unnecessary API calls and impact on performance

gen3/object.py
```diff
@@ -0,0 +1,83 @@
...
+            "authz": authz,
+            "metadata": metadata,
+            "aliases": aliases,
+        }
+        response = requests.post(url, json=body, auth=self._auth_provider)
+        raise_for_status(response)
+        data = response.json()
+        return data["guid"], data["upload_url"]
+
+    def delete_object(self, guid, delete_record=False):
```
paulineribeyre, line 45: There's a bit of a mismatch here, because the logic we came up with for this SDK function is not the same as what the [MDS's "delete object" endpoint](https://github.com/uc-cdis/metadata-service/blob/6214edd/src/mds/objects.py#L390) does (it always deletes from the bucket). And it's a bit weird that this `Gen3Object.delete_object` function is not calling the object deletion endpoint, but the metadata deletion endpoint and then doing its own thing.
I think ideally we should add a parameter to the object deletion endpoint for people who don't want to delete in the bucket, and just call that function here (sorry @nss10). Or at least rename this function so that we can have a `Gen3Object.delete_object` function in the future that just calls the object deletion endpoint.
@Avantol13

gen3/object.py
```diff
@@ -0,0 +1,83 @@
...
+        >>> auth = Gen3Auth(refresh_file="credentials.json")
+        ... file = Gen3File(auth)
+
+    """
+
+    def __init__(self, auth_provider=None):
+        self._auth_provider = auth_provider
+
+    def create_object(self, file_name, authz, metadata=None, aliases=None):
+        url = self.endpoint + "/objects"
```
paulineribeyre, line 33:
```suggestion
        url = self._auth_provider.endpoint + "/objects"
```

gen3/object.py
paulineribeyre, line 45:
```suggestion
    def delete_object(self, guid, delete_file_locations=False):
```

Avantol13: I agree @paulineribeyre , I think we need to update the object deletion endpoint in the metadata-service to behave like we have described here. I honestly didn't remember that endpoint being implemented at all. But if it is, then yeah, we should move this logic there and then this function just calls that endpoint

tests/test_object.py
```diff
@@ -2,48 +2,71 @@
...
+    def _mock_request(url, **kwargs):
+        assert url.endswith("/objects")
+        mocked_response = MagicMock(requests.Response)
+        mocked_response.status_code = 500
+        mocked_response.json.return_value = {"error": "blah"}
+        mocked_response.raise_for_status.side_effect = HTTPError("uh oh")
+        return mocked_response
+
+    requests_mock.side_effect = _mock_request
+    with pytest.raises(Exception):
```
paulineribeyre, line 57: can we use a more specific Exception?

gen3/object.py
```diff
@@ -0,0 +1,54 @@
...
+            "aliases": aliases,
+        }
+        response = requests.post(url, json=body, auth=self._auth_provider)
+        raise_for_status(response)
+        data = response.json()
+        return data["guid"], data["upload_url"]
+
+    def delete_object(self):
+        """
+        Delete the object from indexd, metadata service and optionally all storage locations
```
paulineribeyre, line 45: you're missing the `delete_file_locations` param

tests/test_object.py
```diff
@@ -0,0 +1,72 @@
...
```
paulineribeyre, line 22: can we use a more specific Exception?

gen3/object.py
nss10: Well, thanks to this point,  I found a bug that I almost introduced into the code. I've added that logic in the sdk and also added `unit tests` for this, also got rid of the bug. 🙇

paulineribeyre: i don't see the bugfix but the current version lgtm so i approved 😄

PR #123, opened 2022-03-04
PXP-9276 Add package extraction options

paulineribeyre:
Jira Ticket: [PXP-9276](https://ctds-planx.atlassian.net/browse/PXP-9276)

### New Features
- Package download: add `unpack_packages` flag to enable/disable the unpacking of downloaded packages
- Package download: add `delete_unpacked_packages` flag to allow deleting package files after unpacking them


gen3/tools/download/drs_download.py
```diff
@@ -1276,9 +1330,7 @@ def download_drs_object(
     Returns:
         List of DownloadStatus objects for the DRS object
     """
-    _download_obj(
-        hostname, auth, object_id, output_dir, show_progress
-    )  # pragma: no cover
+    _download_obj(hostname, auth, object_id, output_dir, show_progress)
```
mfshao, line 1333: I think we should put those 2 new arguments into here (and its parent functions as well), so that those `drs-pull` cli cmd can also use them

paulineribeyre: thanks! done

mfshao (approved):
very cool! thanks 👍

PR #131, opened 2022-05-13
Feat/download manifest input

Avantol13:
### New Features
- Download manifest support for providing an input manifest, which will only query indexd for the checksums from that input (instead of downloading all records)

### Breaking Changes
- no longer support python 3.6, 3.7, or 3.8 for a few reasons. pandas no longer supports 3.6 or 3.7 and pandas is required for the core SDK functionality of submitting files (we may replace pandas in the future, but for now this is a limitation). Python >3.8 is required by indexd and some supporting indexdclient libraries going forward (as we're updating all services to 3.9)

### Bug Fixes
- upgrade `black` in pre-commit hooks to avoid issue with newer python versions

### Improvements
- refactor utility functions into utils files

### Dependency updates
- many updates, see diff. Things had not been updated in a while


gen3/cli/objects.py
```diff
@@ -55,8 +55,19 @@ def manifest():
     type=int,
     show_default=True,
 )
+@click.option(
+    "--input-manifest",
```
Avantol13, line 59: this is the real new feature here, an optional argument to the `gen3 objects manifest read --input-manifest foobar.csv`

gen3/index.py
```diff
@@ -202,6 +202,34 @@ async def async_get_records_on_page(self, limit=None, page=None, _ssl=None):
 
         return response.get("records")
 
+    @backoff.on_exception(backoff.expo, Exception, **DEFAULT_BACKOFF_SETTINGS)
+    async def async_get_records_from_checksum(
```
Avantol13, line 206: effectively identical to the async_get_records_on_page above, except with different query params for checksums instead of pages

gen3/tools/indexing/index_manifest.py
```diff
@@ -95,189 +79,6 @@ def __init__(self, processed_files=0, num_total_files=0):
         self.num_total_files = num_total_files
 
 
-def get_and_verify_fileinfos_from_tsv_manifest(
```
Avantol13, line 98: this got relocated (unchanged) into a utils file

gen3/tools/indexing/download_manifest.py
```diff
@@ -96,38 +105,72 @@ async def _write_all_index_records_to_file(
...
+        pages = ",".join(map(str, page_chunks[x])) if page_chunks else ","
+        record_checksums = (
+            ",".join(map(str, record_chunks[x])) if record_chunks else ","
+        )
 
         # call the cli function below and pass in chunks of pages for each process
         command = (
             f"python {CURRENT_DIR}/download_manifest.py --commons_url "
-            f"{commons_url} --pages {pages} --num_processes {num_processes} "
+            f"{commons_url} --pages {pages} --record-checksums {record_checksums} --num_processes {num_processes} "
```
SpencerAxelrod, line 173: `OSError: [Errno 7] Argument list too long`
when I try with genome_file_manifest.tsv, {record_checksums} is extremely long, I believe causing the error. Maybe put the checksums in a separate file so they are not all in one command line args?

Avantol13: shoot, good point, yeah I think I need to use some temporary files

Avantol13: Fixed in newer commits

george42-ctds (approved):
changes to python code look good.

PR #140, opened 2022-07-07
BMH-111

tianj7:
Jira Ticket: [BMH-111](https://ctds-planx.atlassian.net/browse/BMH-111)

### New Features

- Gen3Auth can now communicate with external WTS instances 
- Gen3Auth can obtain access token to target Gen3 commons based on user specified endpoint parameter
- Gen3Auth will set endpoint correctly based on user input


gen3/auth.py
```diff
@@ -311,12 +322,15 @@ def _write_to_file(self, cache_file, content):
...
+    def get_access_token(self, endpoint=None):
         """Get the access token - auto refresh if within 5 minutes of expiration"""
         if not self._access_token:
-            cache_file = token_cache_file(
-                self._refresh_token and self._refresh_token["api_key"] or self._wts_idp
-            )
+            if self._use_wts == True:
+                cache_file = token_cache_file(self._wts_idp)
+            else:
+                cache_file = token_cache_file(
```
Avantol13, line 331: I'm not sure I understand what this token_cache_function is doing. Why are you passing a boolean expression (foo and bar) ?

gen3/auth.py
```diff
@@ -378,3 +392,55 @@ def curl(self, path, request=None, data=None):
...
+    def get_access_token_from_wts(self, endpoint=None):
+        """
+        Try to fetch an access token for the given idp from the wts
+        in the given namespace.  If idp is not set, then default to "local"
+        """
+        # attempt to get a token from the workspace-token-service
+        auth_url = get_wts_endpoint(self._wts_namespace) + "/token/"
+        if self._wts_idp and self._wts_idp != "local":
+            auth_url += "?idp={}".format(self._wts_idp)
+        try:
```
Avantol13, line 405:
```suggestion
        
        try:
```

gen3/auth.py
```diff
@@ -378,3 +392,55 @@ def curl(self, path, request=None, data=None):
...
+            resp = requests.get(auth_url)
+            if (resp and resp.status_code == 200) or (self._external_wts_host == None):
```
Avantol13, line 407: Unless it's vital that the `self._external_wts_host == None` be compared to None, let's just use Python's truthiness. So ```
if (resp and resp.status_code == 200) or not self._external_wts_host:
```

File: gen3/auth.py [hunk 4]
```diff
@@ -378,3 +392,55 @@ def curl(self, path, request=None, data=None):
...
+                return _handle_access_token_response(resp, "token")
+        except Exception as e:
+            if self._external_wts_host == None:
```
Avantol13, line 410: see above re: None

gen3/auth.py
```diff
@@ -378,3 +392,55 @@ def curl(self, path, request=None, data=None):
...
+            providerList = get_wts_idps(self._wts_namespace, self._external_wts_host)
+            matchProviders = list(
+                filter(
+                    lambda provider: provider["base_url"] == endpoint,
+                    providerList["providers"],
+                )
+            )
+            if len(matchProviders) == 0:
+                raise ValueError(
+                    "No idp matched with endpoint value provided. Query /wts/external_oidc/ for list of all available idps."
```
Avantol13, line 434: can you just provide the list of all available idps here in the error message?

gen3/auth.py
```diff
@@ -378,3 +392,55 @@ def curl(self, path, request=None, data=None):
...
+                )
+            elif len(matchProviders) > 1:
+                raise ValueError(
+                    "Multiple idps matched with endpoint value provided. Query /wts/external_oidc/ for specific idp names."
```
Avantol13, line 438: see above

tests/conftest.py
```diff
@@ -148,6 +149,12 @@ def auth_request(*args, **kwargs):
     port = 8001
     settings["auth"].arborist = MockArboristClient()
     indexd = Process(target=run_indexd, args=[port])
+    # Add this line because OS X multiprocessing default is spawn which will cause pickling errors
+    # unstable and not supported on OS X, forking is only supported on Unix
```
Avantol13, line 153:
```suggestion
    # NOTE: fork is unstable and not technically supported on OS X, forking is only supported on Unix
```

tests/test_auth.py
```diff
@@ -177,3 +198,73 @@ def _mock_request(url, **kwargs):
...
+    """
+    try:
+        auth = gen3.auth.Gen3Auth(endpoint="https://caninedc.org", idp="canine-google")
+        # auth object should not be initialized
+        assert not auth
+    except ValueError as e:
+        assert str(e) == "Only one of 'endpoint' and 'idp' can be specified."
+
+
+def test_auth_init_with_external_wts():
```
Avantol13, line 215: I think we need more tests than these. We want to test things like passing in an IdP that doesn't exist, what if token and IdP are different/same

Avantol13 (changes requested):

gen3/auth.py
```diff
@@ -206,6 +202,20 @@ def __init__(self, endpoint=None, refresh_file=None, refresh_token=None, idp=Non
...
+                        == endpoint_from_token(self._refresh_token["api_key"])
+                    ):
+                        self._use_wts = True
+                        self._external_wts_host = (
+                            endpoint_from_token(self._refresh_token["api_key"])
+                            + "/wts/"
+                        )
+                        self.get_access_token(endpoint)
+                except Gen3AuthError as g:
+                    raise g
```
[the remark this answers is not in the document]
tianj7: This block is making sure we raise the correct Gen3AuthError raised (cannot obtain token). The except block below is catching all exceptions and raise a value error indicating the refresh file couldn't be loaded. In reality, the refresh token file could be loaded correctly and the fault is actually not being able to obtain an access token from the local/remote wts instance.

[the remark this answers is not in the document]
Avantol13: oh, I see know. Okay, let's maybe add a log here to clarify that's the error that's occuring

gen3/auth.py
Avantol13: isn't this the function? https://github.com/uc-cdis/gen3sdk-python/blob/19cadb5f8e47b2f8ecff50914a9be04a0a59070f/gen3/auth.py#L91 it looks like the input is just a key to encode. But `_wts_idp` is just a string for the name, it's not a key. 

So are you reusing this same file to cache just the name? it's maybe a little misleading.. perhaps we just need to rename some stuff in the code to make it clean that this is not ONLY a cache file for the key now. 

I also don't know how I feel about the `foo and bar` logic as an argument into the function. The behavior of which is used is understandable but a little too strange at first glance to me.

```
>>> def test(foobar):
...     print(foobar)
... 
>>> test("string" and "anotherstring")
anotherstring
>>> test("string" and None)
None
>>> test(None and "anotherstring")
None
>>> 
```

You could just:
```
if self._refresh_token:
     cache_file = token_cache_file(self._refresh_token.get("api_key", ""))
```
or if you really want a one-liner (which is debatable if it's more readable - I don't think it is) you could do:
```
cache_file = token_cache_file(getattr(self, "_refresh_token", {}).get("api_key", ""))
```

tianj7: Thanks, that makes sense. For the wts case, think I will use `self._refresh_token.get("api_key", "") + _wts_idp ` as the key which should suffice for this purpose.

gen3/auth.py
```diff
@@ -77,25 +77,17 @@ def get_wts_endpoint(namespace=os.getenv("NAMESPACE", "default")):
...
 
-def get_wts_idps(namespace=os.getenv("NAMESPACE", "default")):
-    resp = requests.get(get_wts_endpoint(namespace) + "/external_oidc")
+def get_wts_idps(namespace=os.getenv("NAMESPACE", "default"), external_wts_host=None):
+    wts_url = None
+    if external_wts_host == None:
+        wts_url = get_wts_endpoint(namespace)
+    else:
+        wts_url = external_wts_host
+    resp = requests.get(wts_url + "external_oidc/")
```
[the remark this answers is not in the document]
tianj7: Fixed

gen3/auth.py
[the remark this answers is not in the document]
tianj7: Fixed

gen3/auth.py
```diff
@@ -206,6 +202,20 @@ def __init__(self, endpoint=None, refresh_file=None, refresh_token=None, idp=Non
                         file_data = f.read()
                     self._refresh_token = json.loads(file_data)
                     assert "api_key" in self._refresh_token
+                    # if both endpoint and refresh file are provided, compare endpoint with iss in refresh file
+                    # May need to use network wts endpoint
+                    if (not endpoint == None) and (
```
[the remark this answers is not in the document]
tianj7: the second condition is comparing a ) user supplied endpoint and b) endpoint from the refresh file. If they do NOT agree with each other, then we know we need to use WTS. If they do, then we could just simply load the refresh_file which we have already done. The none reference check has been fixed.

gen3/auth.py
tianj7: fixed

gen3/auth.py
```diff
@@ -270,18 +280,19 @@ def _handle_401(self, response, **kwargs):
...
-                self._wts_namespace, self._wts_idp
-            )
+            self._access_token = self.get_access_token_from_wts(endpoint)
         else:
             self._access_token = get_access_token_with_key(self._refresh_token)
         self._access_token_info = decode_token(self._access_token)
-        cache_file = token_cache_file(
-            self._refresh_token and self._refresh_token["api_key"] or self._wts_idp
-        )
+        if self._use_wts == True:
```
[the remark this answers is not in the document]
tianj7: fixed.

gen3/auth.py
tianj7: fixed.

gen3/auth.py
tianj7: fixed.

gen3/auth.py
tianj7: code refactored per conversation.

gen3/auth.py
tianj7: fixed.

gen3/auth.py
tianj7: fixed.

gen3/auth.py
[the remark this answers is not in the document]
tianj7: the trailing "/" in "/external_oidc/" is required to get the correct json return value.

tests/test_auth.py
tianj7: Added new unit tests

gen3/auth.py
```diff
@@ -378,3 +401,89 @@ def curl(self, path, request=None, data=None):
...
+                logging.warning(
+                    "Accessing local WTS did not work. Trying network call since external WTS url is available"
+                )
+
+        # local workspace wts call failed, try using a network call
+        # First get access token with WTS host
+        wts_token = get_access_token_with_key(self._refresh_token)
+        auth_url = self._external_wts_host + "token/"
+
+        providerList = get_wts_idps(self._wts_namespace, self._external_wts_host)
```
Avantol13, line 432:
```suggestion
        provider_list = get_wts_idps(self._wts_namespace, self._external_wts_host)
```

python locally scoped variables should be `lowercase_underscore_separated`

gen3/cli/objects.py
```diff
@@ -257,13 +257,6 @@ def objects_manifest_validate_format(
     default="indexing-output-manifest.csv",
     show_default=True,
 )
-@click.option(
-    "--force-metadata-columns-even-if-empty",
```
Avantol13, line 261: I think a merge went wrong? This is a new feature that shouldn't be removed

gen3/tools/indexing/index_manifest.py
```diff
@@ -374,35 +367,6 @@ def index_object_manifest(
         manifest_file_delimiter(str): manifest's delimiter
         output_filename(str): output file name for manifest
         submit_additional_metadata_columns(bool): whether to submit additional metadata to the metadata service
-        force_metadata_columns_even_if_empty(bool): force the creation of a metadata column
```
Avantol13, line 377: all of this is the new feature that should still exist, I think a merge went wrong and you got rid of recent commits on master

tests/test_auth.py
```diff
@@ -177,3 +198,338 @@ def _mock_request(url, **kwargs):
...
+    try:
+        auth = gen3.auth.Gen3Auth(endpoint="https://caninedc.org", idp="canine-google")
+        # auth object should not initialize successfully
+        assert not auth
+    except ValueError as e:
+        assert str(e) == "Only one of 'endpoint' and 'idp' can be specified."
+
+
+def test_auth_init_with_matching_endpoint_and_refresh_file():
+
```
Avantol13, line 216: docstring

tests/test_auth.py
```diff
@@ -177,3 +198,338 @@ def _mock_request(url, **kwargs):
...
+    try:
+        with open("testCred.json", "w") as f:
+            json.dump(test_key, f)
+    except Exception as e:
+        print(e)
```
Avantol13, line 221: why are we catching and printing an exception in the unit test?

tests/test_auth.py
```diff
@@ -177,3 +198,338 @@ def _mock_request(url, **kwargs):
...
+    auth = gen3.auth.Gen3Auth(endpoint=test_endpoint, refresh_file="testCred.json")
+    assert auth._use_wts == False
+    assert auth.endpoint == test_endpoint
+    os.remove("testCred.json")
```
Avantol13, line 225: is this safely done even if that file doesn't exist?

tests/test_auth.py
```diff
@@ -177,3 +198,338 @@ def _mock_request(url, **kwargs):
...
+                print(e)
+
+            try:
+                auth = gen3.auth.Gen3Auth(
+                    endpoint=test_external_endpoint, refresh_file="testCred.json"
+                )
+                # auth object should not initialize successfully
+                assert not auth
+            except ValueError as e:
+                assert str(e).startswith(
```
Avantol13, line 297: same comment as above, we should just test for the ValueError, not the content of the message

Avantol13: it will make the tests a little too brittle for my liking

tests/test_manifests.py
```diff
@@ -764,79 +764,6 @@ def test_index_manifest_additional_metadata(gen3_index, gen3_auth):
     assert mds_records[guid] == {"fancy_column": "fancy_data"}
 
 
-@pytest.mark.parametrize("force_metadata_columns_even_if_empty", [True, False])
-def test_index_manifest_additional_metadata_force(
```
Avantol13, line 768: this looks like another part of the bad merge, we don't want to remove this

tests/test_manifests.py
```diff
@@ -63,7 +63,6 @@ def test_verify_manifest(mock_index):
     """
     Test that verify manifest function correctly writes out log file
     with expected error information.
-
```
Avantol13, line 66: seems unnecessary

Avantol13 (approved):

PR #144, opened 2022-07-15
BMH-163

tianj7:
Jira Ticket: BMH-163
### Improvements
- Change drs download mds call to use default python request.raise_for_status() instead of gen3utils version.
- renamed gen3utils raise_for_status() to raise_for_status_and_print_error() for clarity. Refactored code.


gen3/tools/download/drs_download.py
```diff
@@ -1021,7 +1021,7 @@ def download(
             ext = os.path.splitext(entry.file_name)[-1]
             if unpack_packages and ext in PACKAGE_EXTENSIONS:
                 try:
-                    mds_entry = self.metadata.get(entry.object_id)
+                    mds_entry = self.metadata.get(entry.object_id, no_throw=True)
```
mfshao, line 1024: I don't understand why we need to do this

The temporary mitigation solution is to use `--no-unpack-packages` to bypass this check
The correct fix for this should be: if Gen3SDK is being used in a Gen3 ecosystem workspace, it should figure out which child MDS to query and send the query to the correct commons to get metadata back

Disabling throwing exception from here makes no sense, and it is actually going to mute errors that should have been thrown out if SDK is being used in a standlone envirnmont, for example, by users who is going to use SDK on their local computers

mfshao (changes requested):

gen3/tools/download/drs_download.py
tianj7: To my understand, the objects in the BRH notebook user is trying to pull are zip files and not data packages. They do not need to be unpacked but drs download will by default attempt to contact mds. There is no associated metadata for these objects so 404 warnings will appear, which is confusing to users. The relevant conversation thread is here: https://cdis.slack.com/archives/C03A7E397E2/p1657216612023659

gen3/metadata.py
```diff
@@ -272,23 +272,21 @@ async def async_get(self, guid, _ssl=None, **kwargs):
         return response
 
     @backoff.on_exception(backoff.expo, Exception, **DEFAULT_BACKOFF_SETTINGS)
-    def get(self, guid, **kwargs):
+    def get(self, guid, no_throw=False, **kwargs):
```
Avantol13, line 275: I don't like the naming here. We need to be super explicit about what we're ignoring/forcing in this variable name and subsequent docstring

gen3/tools/download/drs_download.py
Avantol13: I don't have all the context, but I agree that ignoring errors from the MDS call is dangerous. This also ignores legitimate errors that might occur. Can this higher-level function just catch the MDS 404 error and handle that here if you need to? 

The 404 is legitimate in this situation. Is the whole thing failing because of those errors or is it just that the log file ends up with 404s? If it's the latter, I'd argue we add some clarity in a future log message, not suppress the HTTP responses being logged. Like at the end, if you know there was an object with no MDS records, you can print a log at the end like: `Files successfully downloaded. Some files did not have associated metadata (see errors in the logs for more details). This did **not** affect the download of the actual files and may or may not be of concern.` or something

mfshao (commented):
changes are being discussed

Avantol13 (approved):

Avantol13 (approved):

PR #147, opened 2022-07-25
BMH-163 & BMH-189

tianj7:
### Improvements
- Add warning message when WTS token could not be obtained
- Hide metadata call 404 error when attempting to unpack packages and only show custom logs instead.
- Fix backoff behavior to not retry after receiving 404 error.
- Disable Backoff default logger and only show custom logs


george42-ctds (dismissed):
looks good to me!

MichaelLukowski (approved):
lgtm

PR #151, opened 2022-08-03
feat(crosswalk): publish crosswalk initial support, tests, and docs

Avantol13:
This is everything to support a CLI and SDK function for publishing crosswalk metadata from local CSVs. There is a comprehensive new docs page describing all the details. See https://github.com/uc-cdis/gen3sdk-python/blob/ca93fd016c66cfe00b85b54ec04242f9fb7bb6bc/docs/howto/crosswalk.md

```
$ gen3 objects crosswalk publish --help
Usage: gen3 objects crosswalk publish [OPTIONS] FILE MAPPING_METHODOLOGY

  Publishes specified crosswalk from local files to Gen3 instance.

  FILE

      A file path to a CSV containing mapping identifiers. Has a specialized
      column naming format. Column names MUST be pipe-delimited and contain:

              commons url | identifier type | identifier name

      You can have any number of columns for mapping.

  MAPPING_METHODOLOGY

      A string description of how the mapping in the provided file was
      obtained.

Options:
  --info PATH  File with an additional description about the identifiers in
               the crosswalk file. The format is a CSV with columns: commons
               url,identifier name,description. Be sure to use the same
               commons url and identifier name used in the crosswalk file.

  --help       Show this message and exit.
```

and 

```
$ gen3 -vv objects crosswalk read --help
Usage: gen3 objects crosswalk read [OPTIONS]

  Reads crosswalk data from Gen3 instance into local files.

Options:
  --output-file PATH  filename for output  [default: crosswalk.csv]
  --help              Show this message and exit.

```

### New Features
- publish crosswalk initial support, tests, and docs


gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+CROSSWALK_NAMESPACE = "crosswalk"
+
+logging = get_logger("__name__", log_level="debug")
+
+
+async def read_crosswalk_metadata(
+    auth, output_filename="crosswalk_metadata.csv", limit=500
+):
+    """
+    fetch crosswalk metadata from a commons and output to crosswalk-metadata.tsv
```
paulineribeyre, line 36:
```suggestion
    fetch crosswalk metadata from a commons and output to crosswalk-metadata.csv
```

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+
+            for metadata_line in metadata_info_reader:
+                # unique key of 2 first columns, value is the description (last column)
+                crosswalk_info[
+                    metadata_line["commons_url"].strip()
+                    + "|"
+                    + metadata_line["identifier_name"].strip()
+                ] = metadata_line["description"].strip()
+
+        logging.debug(f"crosswalk_info: {crosswalk_info}")
```
paulineribeyre, line 207: there's a handful of debug logs that i'm not sure you left on purpose

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+
+            # refresh list if needed, then get a guid
+            if not guids_available_for_use:
+                guids_available_for_use = index.get_valid_guids(count=1000)
+            guid = guids_available_for_use.pop(-1)
+
+            logging.info(f"crosswalk metadata for {guid}: {final_metadata}")
+
+            # call update with merge to ensure this doesn't wipe out any
+            # non-crosswalk namespaced blocks of
```
paulineribeyre, line 338: of?

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+                ]
+
+            if len(pending_requests) == MAX_CONCURRENT_REQUESTS:
+                await asyncio.gather(*pending_requests)
+                pending_requests = []
+
+        await asyncio.gather(*pending_requests)
+
+
+def try_delete_crosswalk_guid(auth, guid):
```
paulineribeyre, line 357: `try_delete_crosswalk_guid` and `_sanitize_tsv_row` not used?

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+    "escapechar": "\\",
+}
+
+GUID_TYPE = "subject"
+CROSSWALK_NAMESPACE = "crosswalk"
+
+logging = get_logger("__name__", log_level="debug")
+
+
+async def read_crosswalk_metadata(
```
paulineribeyre, line 32: this function is hard to understand when you're not familiar with the crosswalk data and could use some comments

gen3/metadata.py
```diff
@@ -120,7 +119,7 @@ def is_healthy(self):
             response = requests.get(
                 self.endpoint + "/_status", auth=self._auth_provider
             )
-            raise_for_status_and_print_error(response)
+            response.raise_for_status()
```
paulineribeyre, line 122: why not keep `raise_for_status_and_print_error`? `raise_for_status` does not output the human friendly error messages, just the codes

gen3/metadata.py
```diff
@@ -365,6 +365,8 @@ async def async_create(
             overwrite (bool, optional): whether or not to overwrite existing data
             _ssl (None, optional): whether or not to use ssl
         """
+        aliases = aliases or []
+        # todo handle aliases
```
paulineribeyre, line 369: should this be implemented before merging this PR or is it a backlog item?

gen3/metadata.py
```diff
@@ -416,8 +420,11 @@ async def async_update(self, guid, metadata, _ssl=None, **kwargs):
                 attached to the provided GUID as metadata
             _ssl (None, optional): whether or not to use ssl
         """
+        # TODO handle aliases
```
paulineribeyre, line 423: same here

docs/howto/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+      "mapping_methodologies": [
+        "NHLBI provided a file of subject IDs for the PETAL study that directly associate a PETAL ID with a BDCat Subject Identifier.",
+        "A Linkage Honest Broker provided MIDRC with what Masked N3C IDs match MIDRC cases via a system-to-system handoff."
+      ]
+    }
+  }
+}
+```
+
+## Crosswalk Data Upload Using Gen3 SDK/CLI
```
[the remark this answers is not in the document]
Avantol13: ooh, good point, yeah it should merge

docs/howto/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+Gen3 SDK Command
+
+```
+gen3 objects crosswalk publish crosswalk_2.csv -m "A Linkage Honest Broker provided MIDRC with what Masked N3C IDs match MIDRC cases via a system-to-system handoff." --info crosswalk_optional_info_2.csv
+```
+
+## Gen3 SDK handling crosswalk.csv submission to MDS
+
+```
+gen3 objects crosswalk read/publish/verify/delete
```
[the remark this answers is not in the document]
Avantol13: oh yeah these are rough notes I had and never actually updated to the final command. I'll adjust

gen3/cli/objects.py
```diff
@@ -324,3 +334,97 @@ def objects_manifest_delete_all_guids(ctx, file):
...
 manifest.add_command(objects_manifest_publish, name="publish")
 manifest.add_command(objects_manifest_delete_all_guids, name="delete-all-guids")
+
+
+@click.command(
+    help=(
+        """
+    Publishes specified crosswalk from local files to Gen3 instance.\n
+
+    FILE\n
```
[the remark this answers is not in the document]
Avantol13: click doesn't allow specifying help on arguments. it's super unintuitive, I know. but they're following some precedent BASH or SH put forward. `Arguments cannot be documented this way. This is to follow the general convention of Unix tools of using arguments for only the most necessary things and to document them in the introduction text by referring to them by name.` https://click.palletsprojects.com/en/4.x/documentation/ trying to do it raises an exception

gen3/metadata.py
paulineribeyre: in HEAL the confusion was because of a specific function that tries to get something from MDS, where getting a 404 would be normal. it was fixed in #144

it's more useful in function where getting an unsuccessful response is not normal

> calling functions should catch exceptions and log appropriate text

if the called function uses `raise_for_status` and fails, i'm not sure the calling function can retrieve the error details form the response?

Avantol13: calling function should be able to, this just raises an HTTPError exception that can be caught, handled, printed, or additional logs added. To your point though, I'm not sure that the response.text gets dumped into the HTTPError.reason but it should... https://docs.python.org/3/library/urllib.error.html#urllib.error.HTTPError 

after some researcher, I think the whole response gets attached https://github.com/psf/requests/blob/d09659997cd1e3eca49a07c59ece5557071c0ab9/requests/exceptions.py#L22 so they should have the same information through the HTTPError

I believe the fix in #144 removed the print for `response.raise_for_status()` in the single location where the problem existed. This PR changes all functions in the metadata file to have that same behavior.

And now I'm remembering that this uses `print` and not logging, so it doesn't allow the caller to set log levels to ignore stuff they don't want to see.

gen3/metadata.py
Avantol13: backlog, there is no alias support in MDS right now, that's the next thing I'm working on. I wanted to get the base functionality reviewed and merged and then go back and add this since it should be small

gen3/tools/metadata/crosswalk.py
Avantol13: :+1: will do

gen3/tools/metadata/crosswalk.py
Avantol13: I did leave on purpose, more debug logs the better if there's an issue in the future, was my thought

gen3/tools/metadata/crosswalk.py
Avantol13: lol, you can see how I got distracted in the middle of writing a sentence :face_with_spiral_eyes:

gen3/tools/metadata/crosswalk.py
Avantol13: delete will get used once I add the `delete` command, I'll add a TODO

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+                                ]
+                            )
+                            crosswalk_columns.add(column_name)
+
+                            crosswalk_info[
+                                commons_url + "|" + identifier_name
+                            ] = indentifer_info.get("description")
+
+                    json.dump(guid_crosswalk_metadata, cached_guid_file)
+                    all_fields |= set(guid_crosswalk_metadata.keys())
```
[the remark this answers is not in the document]
Avantol13: it's not, whoops. Went through some iterations here. Will clean it up

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+                with open(
+                    f"{metadata_cache_dir}/{guid.replace('/', '_')}", "w+"
+                ) as cached_guid_file:
+                    guid_crosswalk_metadata = guid_metadata[CROSSWALK_NAMESPACE]
+
+                    for commons_url, commons_crosswalk in guid_crosswalk_metadata.get(
+                        GUID_TYPE
+                    ).items():
+                        # don't interpret mapping info as a column in the crosswalk file
+                        if commons_url == "mapping_methodologies":
```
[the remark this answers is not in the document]
Avantol13: non-crosswalk columns should be allowed to be merged in, but this file differs in structure from the traditional tsv for metadata so I didn't want to combine the 2 concepts. I'll add more comments to make all this clearer though

gen3/tools/metadata/crosswalk.py
```diff
@@ -0,0 +1,388 @@
...
+                                    identifier_name,
+                                ]
+                            )
+                            crosswalk_columns.add(column_name)
+
+                            crosswalk_info[
+                                commons_url + "|" + identifier_name
+                            ] = indentifer_info.get("description")
+
+                    json.dump(guid_crosswalk_metadata, cached_guid_file)
```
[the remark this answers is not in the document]
Avantol13: except I think I'm still just storing everything I need in memory :face_in_clouds: this might be overkill, let me take a closer look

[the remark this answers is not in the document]
Avantol13: oh wait, no I just keep the columns and descriptions in memory, dump the actual metadata in the file, then read it later. Okay I confused myself with this so I'll definitely add some more comments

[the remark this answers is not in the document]
paulineribeyre: oh i see, my bad

paulineribeyre (approved):

PR #161, opened 2022-10-05
Gen3Index fixes

Avantol13:
correct some issues, ensure params are passed, ensure update calls are successfully constructed

### Bug Fixes
- Ensure params are passed correctly for manifest publishing
- Ensure calls to API are constructed correctly for creating and updating records (`Gen3Index.async_create_record` and `Gen3Index.async_update_record`)
- Fix `backoff` max number of tries override with `GEN3SDK_MAX_RETRIES` environment variable


gen3/index.py
```diff
@@ -736,15 +738,29 @@ async def async_update_record(
...
+            del record["baseid"]
+            del record["size"]
+            del record["hashes"]
+            del record["did"]
+
+            logging.info(f"PUT-ing record: {record}")
+
             async with session.put(
-                f"{self.client.url}/index/{guid}/rev={revision}",
+                f"{self.client.url}/index/{guid}?rev={revision}",
```
paulineribeyre, line 763: oof i can't believe this has been broken for this long

paulineribeyre (dismissed):

gen3/index.py
Avantol13: :sweat_smile: me either

gen3/index.py
```diff
@@ -473,30 +473,39 @@ async def async_create_record(
...
                 "hashes": hashes,
                 "size": size,
-                "file_name": file_name,
-                "metadata": metadata,
-                "urls_metadata": urls_metadata,
-                "baseid": baseid,
-                "acl": acl,
-                "authz": authz,
-                "version": version,
+                "urls": urls or [],
```
BinamB, line 479: isn't `urls` a required field for indexing? If its not available then it shouldn't be a valid manifest right?

thanh-nguyen-dang (approved):

gen3/index.py
paulineribeyre: it's required to provide a list, but it accepts an empty list, so i think that counts as optional? 😄 this isn't for indexing, it's to index a single record, and there are a lot of use cases for indexing before having the URL

BinamB: gotcha

BinamB (approved):

PR #162, opened 2022-10-07
Feat/update disco metadata

mfshao:
Add option arg `update_registered_metadata` (defaults to `True`) to `publish_discovery_metadata()` to allow update existing registered discovery metadata without resetting it

`reset_unregistered_metadata` will take precedence over `update_registered_metadata`, which means:
- if `reset_unregistered_metadata == True`, no matter what value does `update_registered_metadata` has, all registered discovery metadata will be reset back to unregistered metadata
- if `reset_unregistered_metadata == False` and `update_registered_metadata == True`, registered metadata will be updated but not reset
- if `reset_unregistered_metadata == False` and `update_registered_metadata == False`, registered metadata will be skipped for update

### Improvements
- allow `publish_discovery_metadata()` to update existing registered discovery metadata without resetting it


cgmeyer (approved):
🍻 🍕

PR #169, opened 2022-12-09
PXP-9373: Unit test for diff manifest

MaribelleHGomez:
Jira Ticket: [PXP-9373](https://ctds-planx.atlassian.net/browse/PXP-9373)

### New Features
- unit test added for the manifest differential functionality

### Deployment changes

[PXP-9373]: https://ctds-planx.atlassian.net/browse/PXP-9373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gen3/tools/diff.py
```diff
@@ -108,46 +109,53 @@ def _precheck_manifests(
...
+
+                manifest_content.append(content)
+
+        return {
+            "csvdict": manifest_content,
+            "headers": [x for x in headers[0] if x in headers[1]],
+        }
+
+    except Exception as e:
+        print(e)
```
BinamB, line 158: maybe you should return e?

tests/test_diff.py
```diff
@@ -0,0 +1,124 @@
...
+        },
+    )
+
+
+def test_file_input_mismatch():
+    """
+    Test for fail due to different file types.
+    """
+
+    with pytest.raises(SystemExit):
```
BinamB, line 81: I suppose I dont get this. I think it should be an assertion. Something that looks like 
```
assert manifest_diff(...) == error
```

tests/test_diff.py
```diff
@@ -0,0 +1,124 @@
...
+        manifest_diff(
+            files=[
+                "tests/test_data/manifest1.csv",
+                "tests/test_data/diff_manifests/manifest3.tsv",
+            ],
+        )
+
```
BinamB, line 88: I think maybe add one more test that tests the same file and shows that there are no differences between the files.

tests/test_diff.py
MaribelleHGomez: If the files are two different types it exits and stops, what kind of error were you thinking?

frankliuao: My $0.02:
I'd thought that Gen3 would have some custom Exceptions but if not, would a `ValueError` be more appropriate than `TypeError`? The arguments for files might both be in strings? It's the read-in file contents that don't match. My naive thoughts.

frankliuao:
Following up on this @BinamB @MaribelleHGomez

BinamB (approved):
lgtm! 🚀

PR #178, opened 2023-03-13
Feat/dbgap fhir

Avantol13:
### New Features
- Class for parsing NIH's dbGaP FHIR API (added as an optional "extra" installation)
- Support in CLI & SDK for combining arbitrary additional metadata with Gen3 Discovery Metadata


gen3/tools/metadata/discovery.py
```diff
@@ -122,6 +134,158 @@ async def output_expanded_discovery_metadata(
...
+        f"metadata's '{discovery_column_to_map_on}' column. Prefixing added "
+        f"metadata with: '{metadata_prefix}'. Output will be: {output_filename}"
+    )
+
+    # First we need to add a prefix if it was provided
+    temporary_prefixed_filename = (
+        CURRENT_DIR + "temp_" + os.path.basename(metadata_file_to_combine)
+    )
+    delimiter = get_delimiter_from_extension(metadata_file_to_combine)
+    with open(metadata_file_to_combine) as metadata_file:
```
paulineribeyre, line 177: can we skip reading the whole file here if no prefix was provided?

gen3/tools/metadata/discovery.py
```diff
@@ -122,6 +134,158 @@ async def output_expanded_discovery_metadata(
...
+                prefixed_metadata_file, fieldnames=new_headers, delimiter=delimiter
+            )
+
+            # write the new header out
+            writer.writeheader()
+
+            for row in reader:
+                writer.writerow(
+                    {
+                        metadata_prefix + key: value
```
paulineribeyre, line 195: i feel like adding a prefix could be a `merge_guids_into_metadata` parameter, instead of reading through the file an extra time here

gen3/nih.py
```diff
@@ -0,0 +1,597 @@
...
+    def _remove_unecessary_fields(self, all_data):
+        # remove some fields deemed unnecessary or not typically populated
+        for field in self.unnecessary_fields:
+            try:
+                del all_data[field]
+            except KeyError:
+                pass
+
+    @staticmethod
+    def _captialize_top_level_keys(all_data):
```
paulineribeyre, line 524:
```suggestion
    def _capitalize_top_level_keys(all_data):
```

gen3/nih.py
```diff
@@ -0,0 +1,597 @@
...
+        # and replace with our parsing
+        del non_extension_data["extension"]
+        all_data.update(non_extension_data)
+
+        # simplify and clean up everything
+        simplified_data = dbgapFHIR._get_simplified_data(fhir_object)
+        all_data.update(simplified_data)
+        self._flatten_relevant_fields(all_data)
+        self._remove_unecessary_fields(all_data)
+        self._captialize_top_level_keys(all_data)
```
paulineribeyre, line 376: should the capitalization be an optional step if we want this module to be generic? or maybe it belongs in the data ingestion script and not the sdk?

gen3/nih.py
```diff
@@ -0,0 +1,597 @@
...
+        try:
+            del dictionary["Content"]
+        except KeyError:
+            pass
+
+        if dictionary.get("StudyConsents", []):
+            for item in dictionary["StudyConsents"]:
+                consents = item.get("StudyConsent")
+                if consents:
+                    dictionary["StudyConsents"] = consents
```
paulineribeyre, line 485: this assumes there's only 1 item in the `dictionary["StudyConsents"]` list

tests/test_discovery.py
```diff
@@ -188,3 +189,45 @@ async def mock_async_create_metadata(guid, metadata, *_, **__):
...
+    """
+    Test the underlying logic for combining metadata manifests.
+    """
+    current_discovery_metadata_file = (
+        "tests/merge_manifests/discovery_combine/discovery.tsv"
+    )
+    metadata_filename = "tests/merge_manifests/discovery_combine/metadata_file.tsv"
+    discovery_column_to_map_on = "guid"
+    metadata_column_to_map = "Id"
+    output_filename = "test_combined_discovery_metadata.tsv"
```
paulineribeyre, line 204: we should delete this output file after the test

tests/test_nih.py
```diff
@@ -0,0 +1,139 @@
...
+    MOCK_NIH_DBGAP_FHIR_RESPONSE_FOR_PHS000007,
+    MOCK_NIH_DBGAP_FHIR_RESPONSE_FOR_PHS000166,
+)
+
+
+def test_dbgap_fhir():
+    """
+    Test DOI creation under various inputs, ensure it still works even when
+    special characters are used. Ensure it doesn't work when invalid input is
+    provided (like not specify prefix OR id).
```
paulineribeyre, line 24: bad copy/paste?

gen3/cli/discovery.py
```diff
@@ -83,6 +84,84 @@ def discovery_read(ctx, limit, agg):
...
+    "output_filename",
+    help="filename for final combined output",
+    default="combined_discovery_metadata.tsv",
+    show_default=True,
+)
+@click.option(
+    "--limit",
+    "limit",
+    help="max number of metadata records to fetch",
+    default=500,
```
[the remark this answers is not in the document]
Avantol13: Yeah honestly I don't think I originally added this. I believe it was to limit responses for agg mds support. I'll take a closer look

gen3/tools/metadata/discovery.py
```diff
@@ -122,6 +134,158 @@ async def output_expanded_discovery_metadata(
...
+    custom_manifests_mapping_config[
+        "indexing_manifest_column_name"
+    ] = discovery_column_to_map_on
+
+    # by default, the functions for parsing the manifests and rows assumes a 1:1
+    # mapping. There is an additional function provided for partial string matching
+    # which we can use here.
+    custom_manifest_row_parsers[
+        "guids_for_manifest_row"
+    ] = get_guids_for_manifest_row_partial_match
```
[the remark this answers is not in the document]
Avantol13: b/c there are cases where there are not 1:1 matches in metadata/discovery. The FHIR example is that we have the FHIR ResearchStudy ID which is something like `phs000123` but in Discovery we have the full phs id `phs000123.v1.p1.c1`. and we want the FHIR metadata to match to ALL consent groups (so including `phs000123.v1.p1.c2`. Hopefully that makes sense

gen3/tools/metadata/discovery.py
Avantol13: I want to avoid modifying that existing behavior as much as possible to avoid potentially breaking the existing functionality. Since this is dealing with Discovery metadata, I don't think this will ever be on a scale where another file iteration greatly affects the overall time (since this is probably going to be dealing with hundreds, not millions, of rows (like indexing does).

gen3/nih.py
```diff
@@ -0,0 +1,597 @@
...
+                "security",
+            ]
+        )
+
+        # per docs, .elementProperties() returns:
+        #       Returns a list of tuples, one tuple for each property that should be
+        #       serialized, as: ("name", "json_name", type, is_list, "of_many", not_optional)
+        for item in fhir_object.elementProperties():
+            variables_to_simplify.add(item[0])
+            variables_to_simplify.add(item[1])
```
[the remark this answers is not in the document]
Avantol13: for some reason, sometimes the name in the json output and the name on the FHIR python object can be different. Since  the final output contains a combination of manually parsed objects and some raw .to_JSON I opted to include both to ensure we get every combination of the name.

However... I guess I am only simplifying the FHIR object here. I can try removing and see if it works

gen3/nih.py
```diff
@@ -0,0 +1,597 @@
...
+                    simple_representation
+                    or dbgapFHIR._get_simple_text_from_fhir_object(
+                        getattr(item, "coding", []), "coding"
+                    )
+                )
+                if simple_representation:
+                    output.append(simple_representation)
+        except TypeError as exc:
+            # ignore non lists unless it's a FHIR object we know how to parse
+            if type(getattr(fhir_object, list_variable, None)) == FHIRReference:
```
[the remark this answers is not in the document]
Avantol13: we don't catch every possible FHIR object type though so we need logic in the except block no matter what. I guess we could do all this and in the else block do the test to try and iterate over.. but we'd still need a catch in case it's not a list

so it could be:

- if/elif FHIR object types
- else, 
  - try
     - for item in fhir_objecct
   - except, unsupported fhir_object
   
whereas right now we have:
- try
  for item in fhir_objecct
- except
  - if/elif FHIR object types
  - else, unsupported fhir_object
  
  
  They're both kinda gross, if you prefer the other one I can update

gen3/nih.py
Avantol13: It's not optional for the NIH dbGaP FHIR server transformation in my opinion. This is an opinionated ETL of the dbgap FHIR data into a more readable format

gen3/tools/metadata/discovery.py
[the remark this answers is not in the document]
paulineribeyre: 👍 should it be configurable in case we have a simpler use case?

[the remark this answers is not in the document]
Avantol13: Well it should work with partial or exact, so this covers all cases. I guess I could explicitly test the exact match case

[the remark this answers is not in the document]
paulineribeyre: i'm only asking because the `get_guids_for_manifest_row_partial_match` docs said it's a lot slower than the exactly match logic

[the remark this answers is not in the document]
Avantol13: Oh, you're right, I can take another look

tests/test_nih.py
Avantol13: lol, yes, I'll fix it

gen3/nih.py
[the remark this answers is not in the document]
paulineribeyre: i'm probably missing something but i thought it could be:
```
if fhir_object is list:
    for item in fhir_object
elif fhir_object is FHIRReference:
    ...
[elifs other FHIR object types)
else:
    unsupported fhir_object
```
anyways this is fine if it works

paulineribeyre (dismissed):

paulineribeyre (approved):

paulineribeyre (approved):

PR #179, opened 2023-03-20
PXP-10845: Add utilities for dsv file operations and manipulation

MaribelleHGomez:
Jira Ticket: [PXP-10845](https://ctds-planx.atlassian.net/browse/PXP-10845)

### New Features
- Utility file for DSV file manipulation; functions include: collecting file contents into a list, finding file intersection and difference, chunking files, splitting file by headers, merging files

### DIFF
--- a/gen3/tools/file_utils.py
+++ b/gen3/tools/file_utils.py
@@ -1,0 +1,430 @@
+import os
+import csv
+
+from typing import Union
+from cdislogging import get_logger
+
+logging = get_logger("__name__")
+
+
+def _get_directory_file_paths(directory: str):
+    """
+    List all file paths in a directory.
+    Args:
+        directory(str):
+            Path of the directory containing the input manifests. All of the manifests contained in directory are assumed to be in a delimiter-separated values (DSV) format, and that there are no other non-DSV files in directory.
+    Returns:
+        List of files in directory
+    """
+
+    logging.info(f"Iterating over files in {directory} directory")
+
+    files = []
+    if os.path.exists(directory):
+        directory_list = os.listdir(directory)
+    else:
+        raise NotADirectoryError()
+    if directory_list:
+        for file in sorted(directory_list):
+            files.append(os.path.join(directory, file))
+
+        files.sort()
+    else:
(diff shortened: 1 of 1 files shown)


gen3/tools/file_utils.py
```diff
@@ -0,0 +1,430 @@
...
+        ):
+            list_two = move_columns(file_one_headers, file_two)
+        elif file_one_headers != file_two_headers:
+            raise AssertionError("Columns do not have matching headers")
+        else:
+            list_two = file_to_list(file_two, has_headers)
+    else:
+        list_two = file_to_list(file_two, has_headers)
+
+    if not strict:
```
BinamB, line 341: i think you need to handle cases where `filename` and `file_name` are the same thing and this code should be smart enough to figure that out. Luckily there are functions in the sdk that already handles that. Check here. https://github.com/uc-cdis/gen3sdk-python/blob/master/gen3/tools/utils.py
Maybe right after we open up the files we standardize the header formats

gen3/tools/file_utils.py
```diff
@@ -0,0 +1,430 @@
...
+            True when headers present
+        strict(bool):
+            When true, will look for exact match; else, will match lower and uppercase strings and match array of different order
+    Returns:
+        List of lists, returns rows that contain matches between all columns
+    """
+
+    logging.info(f"Taking intersection between {file_one} & {file_two}...")
+
+    list_one = file_to_list(file_one, has_headers)
```
BinamB, line 374: Maybe we should validation for all manifests that the utils processes.

gen3/tools/file_utils.py
```diff
@@ -0,0 +1,430 @@
...
+                str(sorted(eval(column)))
+        for row in list_two:
+            for column in row:
+                column.lower()
+                str(sorted(eval(column)))
+
+    return [row for row in list_one if row in list_two]
+
+
+# TODO
```
BinamB, line 403: Can you create a ticket for this TODO and link it here as well?

BinamB (changes requested):

PR #180, opened 2023-03-27
BRH-314

tianj7:
Jira Ticket: https://ctds-planx.atlassian.net/browse/BRH-314

### Improvements
Add confirmation message for successful DRS downloads


gen3/tools/download/drs_download.py
```diff
@@ -47,7 +47,7 @@
 # package formats we handle for unpacking
 PACKAGE_EXTENSIONS = [".zip"]
 
-logger = get_logger("drs-pull", log_level="warning")
+logger = get_logger("drs-pull", log_level="info")
```
Avantol13, line 50: We shouldn't manually set the log_level here. It should be configurable by the client what log level they want to see. This was recently added to the SDK `gen3 -v` and `gen3 -vv`. You should try removing the log_level entirely from here and update the command call to specify the verbosity. If you want info messages, use `gen3 -v ...`

from `gen3 --help`:
```
  -v                 verbose logs show INFO, WARNING & ERROR logs
  -vv                very verbose logs show DEGUG, INFO, WARNING & ERROR logs

```

Avantol13 (changes requested):

gen3/tools/download/drs_download.py
tianj7: Code modified to use default flags and use click.echo to display download success/error messages

gen3/cli/drs_pull.py
```diff
@@ -139,6 +140,12 @@ def download_object(
         delete_unpacked_packages,
     )
 
+    for drs_object_id in res:
+        if res[drs_object_id].status != "downloaded":
```
Avantol13, line 144:
```suggestion
        if drs_object_id in res and res[drs_object_id].status != "downloaded":
```

gen3/cli/drs_pull.py
```diff
@@ -139,6 +140,12 @@ def download_object(
...
+            click.echo(f"Object {drs_object_id} have not been successfully downloaded.")
```
Avantol13, line 145:
```suggestion
            click.echo(f"Object {drs_object_id} has failed to downloaded.")
```

remove "success" b/c at a glance someone might miss the "not" and think it worked

gen3/cli/drs_pull.py
```diff
@@ -139,6 +140,12 @@ def download_object(
...
+        else:
+            click.echo(f"Object {drs_object_id} have been successfully downloaded.")
```
Avantol13, line 147: Do we need it to echo for every successful download or just if one fails? Seems like this might be too much output on the command line

gen3/tools/download/drs_download.py
```diff
@@ -1049,8 +1049,13 @@ def download(
 
                     if delete_unpacked_packages:
                         filepath.unlink()
-
-            completed[entry.object_id].status = "downloaded" if res else "error"
+            if res:
+                completed[entry.object_id].status = "downloaded"
+                logger.info(
```
Avantol13, line 1054: maybe debug instead of info. This could get pretty noisy

gen3/tools/download/drs_download.py
```diff
@@ -1049,8 +1049,13 @@ def download(
...
+                    f"object {entry.object_id} has been successfully downloaded."
+                )
+            else:
+                completed[entry.object_id].status = "error"
```
Avantol13, line 1058: why a log above for success but not for failure here?

Avantol13 (changes requested):
see comments

gen3/cli/drs_pull.py
tianj7: rewrote this to just show individual failures only and a final success/failure message.

Avantol13 (approved):

PR #183, opened 2023-05-04
feat(dois): dbGaP Parsing for DOI metadata automation

Avantol13:
This is building off the core DOI support that was added previously.

The new functionality description and example is in docs here: https://github.com/uc-cdis/gen3sdk-python/blob/a9918cccacf2e324673f910f54985445808956d1/docs/howto/discoveryMetadataTools.md#automate-doi-creation-for-dbgap-datasets

The original DOI support and info can be found just above that, here: https://github.com/uc-cdis/gen3sdk-python/blob/a9918cccacf2e324673f910f54985445808956d1/docs/howto/discoveryMetadataTools.md#dois-in-gen3-discovery-metadata-and-page-for-visualizing-public-doi-metadata

### New Features
- External interface to dbGaP's Study Registration API
- External interface / abstraction for getting all DOI related metadata from dbGaP's APIs
- High level function for automatically getting all necessary DOI metadata for dbGaP datasets, updating/creating DOIs, and updating Gen3's Metadata API with the necessary information

### Improvements
- Refactor/organization of external metadata interfaces and a new ExternalMetadataSourceInterface class to inherit from


gen3/external/nih/dbgap_fhir.py
```diff
@@ -572,70 +577,3 @@ def _capitalize_top_level_keys(all_data):
 
             if key != capitalized_key:
                 del all_data[key]
-
-
-def get_dbgap_accession_as_parts(phsid):
```
Avantol13, line 577: moved to nih.utils

gen3/discovery_dois.py
```diff
@@ -0,0 +1,282 @@
...
+from gen3.utils import (
+    get_delimiter_from_extension,
+    get_or_create_event_loop_for_thread,
+    is_status_code,
+)
+
+logging = get_logger("__name__")
+
+
+def mint_dois_for_dbgap_discovery_datasets(
```
mfshao, line 20: If I'm reading this right looks like this function can also generate DOIs for non-dbgap datasets? So I'm wondering if I want mint DOIs for some other datasets, should I be expected to create another function simiar to this one and uses `_create_or_update_doi_and_persist()` inside it?

gen3/discovery_dois.py
```diff
@@ -0,0 +1,282 @@
...
+    get_doi_identifier_function,
+    publisher,
+    commons_discovery_page,
+    doi_disclaimer,
+    doi_access_information,
+    doi_access_information_link,
+    doi_contact,
+    use_doi_in_landing_page_url=False,
+    doi_metadata_field_prefix="doi_",
+    datacite_api=DataCite.TEST_URL,
```
mfshao, line 33: minor: instead of passing `DataCite.TEST_URL` or `DataCite.PRODUCTION_URL` directly, would it be better if the underlying `DataCite` class takes a `mode` argument, which can be either `test` or `production` and use that to switch between using these 2 URLs inside `DataCite`?

Avantol13: that's a good idea for the common cases, I do still want to allow overriding the url entirely, but I think I can manage to do both. I'll take a look

gen3/external/nih/dbgap_doi.py
```diff
@@ -0,0 +1,296 @@
...
+
+        Args:
+            ids (List[str]): list of IDs to query for
+
+        Returns:
+            Dict[dict]: metadata for each of the provided IDs (which
+                        are the keys in the returned dict)
+        """
+        logging.info("Getting dbGaP DOI metadata...")
+        logging.debug("Provided ids for dbGaP DOI metadata: {ids}")
```
paulineribeyre, line 85:
```suggestion
        logging.debug(f"Provided ids for dbGaP DOI metadata: {ids}")
```

paulineribeyre (commented):
lgtm

gen3/discovery_dois.py
Avantol13: I took your advice and split things out a lot more. Now there's a GetMetadataInterface of which there's a single child class for dbGaP. For other sources in the future, theoretically you just need to add a new interface class and then all this same code will work

gen3/discovery_dois.py
```diff
@@ -228,6 +370,8 @@ def _create_or_update_doi_and_persist(
 
     doi = DigitalObjectIdentifier.from_datacite_create_doi_response(response)
 
+    # TODO: update `is_available` in `additional_metadata`?
```
paulineribeyre, line 373: do we need to address this before merging?

paulineribeyre (dismissed):

gen3/discovery_dois.py
Avantol13: It's a nice to have feature for later - we have to address how to handle displaying old/new versions of dataset metadata in Discovery eventually, and when we decide how to do that and persist in the MDS, we may want to add some logic to automatically set `is_available` to False for old versions (but we'd need to also add logic to know we're updating to a new version).

tl;dr We don't really have the versioning feature fully figured out enough to actually do this, but I wanted to remember where in the code we'd likely need to do it in the future

paulineribeyre (approved):

PR #184, opened 2023-05-19
HP-1017 Feat/disco update

mfshao:
Jira Ticket: [HP-1017](https://ctds-planx.atlassian.net/browse/HP-1017)

### Improvements
- Allow `output_expanded_discovery_metadata()` to fetch discovery metadata stored under different GUID types
- Allow `output_expanded_discovery_metadata()` to save output into either TSV or JSON file
- Allow `publish_discovery_metadata()` to publish metadata from a JSON file

### DIFF
--- a/gen3/tools/metadata/discovery.py
+++ b/gen3/tools/metadata/discovery.py
@@ -104,9 +104,12 @@ async def output_expanded_discovery_metadata(
             else:
                 break
 
+        output_filename_extension = ".tsv"
+        if output_format == "json":
+            output_filename_extension = ".json"
         output_filename = _create_metadata_output_filename(
-            auth, guid_type, output_filename_suffix
-        ) + (".tsv" if output_format == "tsv" else ".json")
+            auth, guid_type, output_filename_suffix, output_filename_extension
+        )
         if output_format == "tsv":
             output_columns = (
                 ["guid"]
@@ -353,7 +356,7 @@ async def publish_discovery_metadata(
         mds = Gen3Metadata(auth_provider=auth)
 
     if not metadata_filename:
-        metadata_filename = _create_metadata_output_filename(auth, guid_type) + ".tsv"
+        metadata_filename = _create_metadata_output_filename(auth, guid_type)
 
     if not (
         metadata_filename.endswith(".csv")
@@ -555,9 +558,10 @@ def _try_parse(data):
     return ""
 
 
-def _create_metadata_output_filename(auth, guid_type="", suffix=""):
+def _create_metadata_output_filename(auth, guid_type, suffix="", file_extension=".tsv"):
     return (
         "-".join(urlparse(auth.endpoint).netloc.split("."))
-        + (f"-{guid_type}" if guid_type else "")
+        + f"-{guid_type}"
         + (f"-{suffix}" if suffix else "")
+        + file_extension
     )
--- a/gen3/tools/metadata/discovery.py
+++ b/gen3/tools/metadata/discovery.py
@@ -35,11 +35,23 @@ logging = get_logger("__name__")
 
 
 async def output_expanded_discovery_metadata(
-    auth, endpoint=None, limit=500, use_agg_mds=False
+    auth,
+    endpoint=None,
+    limit=500,
+    use_agg_mds=False,
+    guid_type="discovery_metadata",
+    output_format="tsv",
+    output_filename_suffix="",
 ):
     """
     fetch discovery metadata from a commons and output to {commons}-discovery-metadata.tsv
     """
+    if output_format != "tsv" and output_format != "json":
+        print(
+            f"Unsupported output file format {output_format}! Only tsv or json is allowed"
+        )
+        return
+
     if endpoint:
         mds = Gen3Metadata(
             auth_provider=auth,
@@ -52,14 +64,13 @@ async def output_expanded_discovery_metadata(
             service_location="mds/aggregate" if use_agg_mds else "mds",
         )
 
-    count = 0
     with tempfile.TemporaryDirectory() as metadata_cache_dir:
         all_fields = set()
         num_tags = 0
 
         for offset in range(0, limit, MAX_GUIDS_PER_REQUEST):
             partial_metadata = mds.query(
-                "_guid_type=discovery_metadata",
+                f"_guid_type={guid_type}",
                 return_full_metadata=True,
                 limit=min(limit, MAX_GUIDS_PER_REQUEST),
                 offset=offset,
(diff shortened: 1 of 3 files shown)


MichaelLukowski (dismissed):

Avantol13 (changes requested):
Can you update the docs on how to use the new feature and maybe some examples? https://github.com/uc-cdis/gen3sdk-python/blob/master/docs/howto/discoveryMetadataTools.md

gen3/tools/metadata/discovery.py
```diff
@@ -35,11 +35,23 @@
...
+    guid_type="discovery_metadata",
+    output_format="tsv",
+    output_filename_suffix="",
 ):
     """
-    fetch discovery metadata from a commons and output to {commons}-discovery-metadata.tsv
+    fetch discovery metadata from a commons and output to {commons}-{guid_type}.tsv or {commons}-{guid_type}.json
     """
+    if output_format != "tsv" and output_format != "json":
+        print(
```
Avantol13, line 50: you should use logging.error instead of print

Avantol13: and maybe raise an exception instead of empty return

tests/test_discovery.py
```diff
@@ -14,9 +14,11 @@
 )
 
 
-@patch("gen3.tools.metadata.discovery._metadata_file_from_auth")
+@patch("gen3.tools.metadata.discovery._create_metadata_output_filename")
 @patch("gen3.metadata.Gen3Metadata.query")
-def test_discovery_read(metadata_query_patch, metadata_file_patch, gen3_auth):
+def test_discovery_read(
```
Avantol13, line 19: you should add another test for a non "discovery_metadata" guid type to cover all the new features you added. You should probably have 2, one for TSV, one for JSON.

gen3/tools/metadata/discovery.py
mfshao: updated

gen3/tools/metadata/discovery.py
```diff
@@ -91,45 +104,60 @@ async def output_expanded_discovery_metadata(
...
-
-        output_filename = _metadata_file_from_auth(auth)
-        with open(
-            output_filename,
-            "w+",
-        ) as output_file:
-            writer = csv.DictWriter(
-                output_file,
-                **{**BASE_CSV_PARSER_SETTINGS, "fieldnames": output_columns},
+        if output_format == "tsv":
```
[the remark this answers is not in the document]
mfshao: I re-orged the code a bit to elimitnate the repeated if/else blocks. There is actually not many overlaps between the TSV and JSON logics since the JSON logic is so much simpler than the TSV ones. But I can try to further extract those code blocks from both TSV and JSON side and put them into 2 new functions if you think that would be better

docs/howto/discoveryMetadataTools.md
```diff
@@ -20,6 +22,68 @@ So you can choose to use the CLI or write your own Python script and use the SDK
 functions yourself. Generally this provides the most flexibility, at less
 of a convenience.
 
+### Export Discovery Metadata into File
+Gen3's SDK can be used to export discovery metadata from a certain Gen3 environment into a file by using the `output_expanded_discovery_metadata()` function. By default this function will query for metadata with `guid_type=discovery_metadata` for the dump, and export the metadata into a TSV file. User can also specified a different `guid_type` values for this operation, and/or choose to export the metadata into a JSON file. When using TSV format, some certain fields from metadata will be flattened or "jsonified" so that each metadata record can be fitted into one row.
```
Avantol13, line 26:
```suggestion
Gen3's SDK can be used to export discovery metadata from a certain Gen3 environment into a file by using the `output_expanded_discovery_metadata()` function. By default this function will query for metadata with `guid_type=discovery_metadata` for the dump, and export the metadata into a TSV file. User can also specify a different `guid_type` values for this operation, and/or choose to export the metadata into a JSON file. When using TSV format, some certain fields from metadata will be flattened or "jsonified" so that each metadata record can be fitted into one row.
```

docs/howto/discoveryMetadataTools.md
```diff
@@ -20,6 +22,68 @@ So you can choose to use the CLI or write your own Python script and use the SDK
 functions yourself. Generally this provides the most flexibility, at less
 of a convenience.
 
+### Export Discovery Metadata into File
+Gen3's SDK can be used to export discovery metadata from a certain Gen3 environment into a file by using the `output_expanded_discovery_metadata()` function. By default this function will query for metadata with `guid_type=discovery_metadata` for the dump, and export the metadata into a TSV file. User can also specify a different `guid_type` values for this operation, and/or choose to export the metadata into a JSON file. When using TSV format, some certain fields from metadata will be flattened or "jsonified" so that each metadata record can be fitted into one row.
+
+Function args:
+* `auth`: a Gen3Auth object, required
```
Avantol13, line 29: I'd actually rather this level of detail about the args be in the docstring in the code (if it isn't alread), and these docs are more for showing practical usage and example input and maybe output. 

The block of code on how to use it should stay there, maybe just show an example output, remove this args section, and then I'd call it good. Similar comments for the below section

Avantol13 (changes requested):
Okay one final request, can you update the command line to allow specifying JSON output for download?

Maybe `--output-type` with default of whatever it is now

```
Usage: gen3 discovery read [OPTIONS]

  Download the metadata used to populate a commons' discovery page into a
  TSV. Outputs the TSV filename with format {commons-
  url}-discovery_metadata.tsv

Options:
  --limit INTEGER  max number of metadata records to fetch  [default: 500]
  --agg            use aggregate metadata service instead of the metadata
                   service  [default: False]

  --help           Show this message and exit.

```

docs/howto/discoveryMetadataTools.md
mfshao: updated

mfshao:
> Okay one final request, can you update the command line to allow specifying JSON output for download?
> 
> Maybe `--output-type` with default of whatever it is now
> 
> ```
> Usage: gen3 discovery read [OPTIONS]
> 
>   Download the metadata used to populate a commons' discovery page into a
>   TSV. Outputs the TSV filename with format {commons-
>   url}-discovery_metadata.tsv
> 
> Options:
>   --limit INTEGER  max number of metadata records to fetch  [default: 500]
>   --agg            use aggregate metadata service instead of the metadata
>                    service  [default: False]
> 
>   --help           Show this message and exit.
> ```

I updated the CLI cmds for read and publish for adding missing optinal fields but I intentionally left out those HEAL only arguments for the publish command since they have less interests to others

Avantol13:
> 

That's a good point, that's fair. Better to keep the CLI opinionated. Oh the last comment I have is that we should remove any project name references from docs. Even if it is HEAL specific, we should find a generic way to say that. Like "some Gen3 instances do x but the majority do y"

docs/howto/discoveryMetadataTools.md
Avantol13: I think the branch is missing a commit, have you pushed the updates?

mfshao:
> > 
> 
> That's a good point, that's fair. Better to keep the CLI opinionated. Oh the last comment I have is that we should remove any project name references from docs. Even if it is HEAL specific, we should find a generic way to say that. Like "some Gen3 instances do x but the majority do y"

make sense, updated those docstrings to remove project names

Avantol13 (dismissed):

Avantol13 (approved):

PR #189, opened 2023-07-14
fix use before define

mfshao:
### Bug Fixes
- Fixed an error caused by calling a variable before defining it


george42-ctds (approved):
Looks good.

PR #190, opened 2023-07-21
feat(PXP-10777): Utils for cascading authorization

k-burt-uch:
Adds support for querying the dbGaP Study Registration API for child studies.

Updates dbgap_study_registration to include the study's accession in get_metadata_for_ids return.

### New Features
- Adds support for querying the dbGaP Study Registration API for child studies.

### Improvements
- Updates dbgap_study_registration to include the study's accession in get_metadata_for_ids return.


gen3/cli/nih.py
```diff
@@ -0,0 +1,96 @@
...
+        > phs002793.v2.p1: phs002795.v1.p1 phs002796.v1.p1 phs002797.v1.p1 phs002798.v1.p1 phs002794.v2.p1
+
+        gen3 nih dbgap-study-registration get-child-studies "phs002076 phs002793"
+         > phs002076.v2.p1: phs002077.v2.p1
+           phs002793.v2.p1: phs002795.v1.p1 phs002796.v1.p1 phs002797.v1.p1 phs002798.v1.p1 phs002794.v2.p1
+    """
+    result = dbgapStudyRegistration().get_child_studies_for_ids(study_name.split(" "))
+
+    if not result:
+        click.echo(f"No child studies found for {study_name}")
```
Avantol13, line 93: this is a new idea (open to your thoughts), but I added a `--silent` recently to `gen3` with the idea that we should be able to silence any logs that are informational and try to get a consistent (ideally parse-able) click.echo from each command. The idea being, I should be able to pipe the output of a command reliably into another script if I --silent the logs.

So I would prefer if this particular line was an info log and if we gathered the results and returned a consistent format, maybe JSON in a single `click.echo`

So I could `gen3 --silent nih dbgap-study-registration get-child-studies "phs002076 phs002793" | python update_metadata.py` or something

Avantol13 (changes requested):

gen3/cli/nih.py
k-burt-uch: Just to confirm:

- Change this "no child studies found" echo into a log.info
- Change results into JSON 

When silent is used, this would suppress the info log for "no child studies found". 

What should happen to the output of the command when there is a child study when silent is used vs not used? Silent is json and no silent is just the output we have now? Or no change?

Also, whats the reason for using the cli in scripts rather than the class itself? bash scripting?

Avantol13: Yeah, bash. basically allowing the user of the CLI to continue using the output in bash or other commands more reliably if they so choose. 

silent just silences logging.whatever's

click.echo happens regardless of log level (in other words, it's a consistent way to return an output)

so the idea would be to consistently return some JSON structure with click.echo and any other informational messages or errors should use logging

k-burt-uch: I went with this json structure 

```
{
  "phs002076.v2.p1": ["phs002077.v2.p1"],
  "phs002793.v2.p1": ["phs002795.v1.p1", "phs002796.v1.p1", "phs002797.v1.p1", "phs002798.v1.p1"]
}
```

for child studies. 

it is pretty bare - should I make it more descriptive?

Avantol13: nah I think that's fine

Avantol13 (approved):

PR #191, opened 2023-07-27
Fix/doi delimiters

Avantol13:
### Improvements
- delimit lists in DOI citation with & so commas in last,first name don't cause confusion


cgmeyer (approved):
Thanks! 🍻

PR #195, opened 2023-08-22
feat(drs-pull-objects): initial support for supplying a list of DRS I…

Avantol13:
…Ds to download
-->

### New Features
- new `drs-pull objects` command which accepts a list of DRS IDs instead of a single one like `drs-pull object`
- new `--silent` arg to silence all logging and only let `click.echo`'s through to stdout (this enables the ability to ensure a consistent stdout response to pipe real consistent results to other commands, e.g. a `gen3 --silent drs-pull object dg.4503/foobar | jq` will always pipe the JSON output into `jq`


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "gen3"
 homepage = "https://gen3.org/"
-version = "4.20.5"
+version = "4.22.0"
```
nss10, line 4: Why is it not `4.21.0`?

nss10 (approved):

nss10 (approved):

PR #211, opened 2024-02-26
change md5 to sha256 for fips compliance

SpencerAxelrod:


Avantol13 (approved):
this does mean people's with a current cache will miss for everything until they get the new sha256 filenames, but that should be fine.

PR #217, opened 2024-04-11
Fix/metadata missing tags

craigrbarnes:

### Bug Fixes
Fixes an issue where metadata tags are not present.


gen3/tools/metadata/discovery.py
```diff
@@ -106,6 +106,8 @@ async def output_expanded_discovery_metadata(
                 for guid in sorted(os.listdir(metadata_cache_dir)):
                     with open(f"{metadata_cache_dir}/{guid}", encoding="utf-8") as f:
                         fetched_metadata = json.load(f)
+                        if fetched_metadata.get("tags", []) is None:
```
k-burt-uch, line 109:
```suggestion
                        if fetched_metadata.get("tags") is None:
```

gen3/tools/metadata/discovery.py
```diff
@@ -190,9 +192,8 @@ def read_mds_into_cache(
                     guid_discovery_metadata = guid_metadata["gen3_discovery"]
                     json.dump(guid_discovery_metadata, cached_guid_file)
                     all_fields |= set(guid_discovery_metadata.keys())
-                    num_tags = max(
-                        num_tags, len(guid_discovery_metadata.get("tags", []))
-                    )
+                    tgs = guid_discovery_metadata.get("tags", [])
+                    num_tags = max(num_tags, 0 if tgs is None else len(tgs))
```
k-burt-uch, line 196: tgs will never be None since `guid_discovery_metadata.get("tags", [])` has `[]` as the default return if `guid_discovery_metadata["tags"]` is None. 

So tgs will always be an array here. This is effectively the same as the change that existed. I would revert this portion of the changeset.

PR #219, opened 2024-05-08
DCF-1735: Add user security module for internal use 

jacob50231:
### New Features
* Added function that scans user yaml files for security concerns


gen3/tools/indexing/user_security.py
```diff
@@ -0,0 +1,150 @@
...
+                            location = "resources" + "/".join(path)
+                            logging.info(
+                                f"PHSID Found in {location} listed as an open program"
+                            )
+
+        if found == False:
+            logging.info(f"{phsid} not found in open projects")
+
+        continue_input = ""
+        while continue_input.lower() not in ["y", "n"]:
```
BedfordWest, line 111: As discussed earlier, let's set a timeout for this and skip it entirely if there are no instances of the phsid found.

BedfordWest (changes requested):
Great work! Just the one change we discussed during our touchpoint.

gen3/tools/indexing/user_security.py
```diff
@@ -48,72 +52,105 @@ def check_yaml(phsid, path, repo="commons-users", verbose=False):
...
+            if found == False:
+                logging.info(f"{phsid} not found in open projects \n")
+                if verbose == False:
+                    # If verbose is true and no phsid is found we would have already continued
+                    continue
+
+    # If we find anything ask for input (eventually times out)
+    if anything_found:
+        print("Would you like to continue? (y/n): ", end="", flush=True)
+        timeout = 5
```
BedfordWest: 5 seconds feels a bit short to me. Can we make it a bit longer? Maybe 15s?

BedfordWest (changes requested):
Overall looks good! Just one request about the timeout duration.

BedfordWest (approved):

gen3/tools/indexing/user_security.py
```diff
@@ -0,0 +1,188 @@
...
+            else:
+                lines = ", ".join(lines)
+                logging.info(f"{phsid} found in lines {lines} \n")
+
+        logging.info(f"========= Scanning file for {phsid} in open projects =========")
+        found = False
+        # Track down open_data_reader
+        policies = content[auth_meth]["policies"]
+        for i in range(len(policies)):
+            if policies[i]["id"] == "open_data_reader":
```
paulineribeyre, line 89: @BedfordWest @jacob50231 I'm a bit late but since the Gen3 SDK is widely used externally, i think the docstring should document the limitations and assumptions we make in this script, ie that "public access" == "open_data_reader" policy. That's not necessarily true:
- What do we mean by "public"? The "anonymous_policies" group or "all_users_policies" group, or both?
- Those groups can include any policies we want, not just "open_data_reader".
- Using the name "open_data_reader" is a convention that isn't enforced and we use a different name sometimes, even internally.

Example: https://github.com/uc-cdis/commons-users/blob/bc1c4f4/users/midrc/user.yaml#L12-L23

If this is meant to be usable by others, we should update the script (check all policies in the 2 groups). If not, we should at least document the assumptions (or just keep this internal script in a different repo).

From a quick read, other things to document:
- The hardcoded `owner = "uc-cdis"` (since the `path` and `repo` are already parameters, why not this one too?)
- What a "phsid" is (external folks may not know what that is, "resource_name" would be more generic)

Also quick note that unit tests would have been nice, also because the sdk is an external tool

paulineribeyre: Actually this would be a great addition to the gen3users tool!

paulineribeyre: https://ctds-planx.atlassian.net/browse/DCF-1771

PR #220, opened 2024-05-22
HP-1458 add external download

george42-ctds:
JIRA ticket: [HP-1458](https://ctds-planx.atlassian.net/browse/HP-1458)

### New Features

* download data from external repos

### Deployment changes

[HP-1458]: https://ctds-planx.atlassian.net/browse/HP-1458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gen3/tools/download/external_file_download.py
```diff
@@ -0,0 +1,348 @@
...
+        auth: Gen3Auth object for access to WTS
+        external_file_metadata (dict):
+        retrievers (dict):
+        download_path: path to write data files
+    Returns:
+        list of download status for each external_file_metadata object, None if errors
+    """
+    completed = {}
+
+    # loop over external file objects
```
george42-ctds, line 137: The `pull_files` will be refactored to pass a list of all the external_file_metadata objects (per retriever type) to the retriever instead of iterating over each item.

gen3/tools/download/external_file_download.py
```diff
@@ -0,0 +1,256 @@
...
+
+REQUIRED_EXTERNAL_FILE_METADATA_KEYS = ["external_oidc_idp", "file_retriever"]
+
+logger = get_logger("__name__", log_level="debug")
+
+
+def download_files_from_metadata(
+    hostname: str,
+    auth: Gen3Auth,
+    metadata_file: str,
```
mfshao, line 26: I feel this function should take metadata input in a form of dict, which means a JSON object. 
And in the future if we need to create a cli cmd of this, then it could take a file path to metadata file as input

gen3/tools/download/external_file_download.py
```diff
@@ -0,0 +1,256 @@
...
+        dictionary of DownloadStatus for external file objects, None if errors
+    """
+    completed = {}
+
+    metadata_grouped_by_retriever = check_data_and_group_by_retriever(
+        file_metadata_list
+    )
+    logger.debug(f"Grouped metadata = {metadata_grouped_by_retriever}")
+
+    for key, val in metadata_grouped_by_retriever.items():
```
mfshao, line 133: `metadata_grouped_by_retriever` could be `None` from L128 above
I think we need to either check for this edge case or make it returns `{}` instead of `None`

gen3/tools/download/external_file_download.py
```diff
@@ -0,0 +1,256 @@
...
+
+    download_status = pull_files(
+        wts_server_name=hostname,
+        auth=auth,
+        file_metadata_list=external_file_metadata,
+        retrievers=retrievers,
+        download_path=download_path,
+    )
+
+    for key in download_status:
```
mfshao, line 101: According to L159 `download_status` can be `None` here

docs/howto/externalFileDownloading.md
```diff
@@ -0,0 +1,39 @@
...
+}
+```
+
+The `external_oidc_idp` field is required. It is used to determine how to get a token from the `workspace token service`.
+
+The `file_retriever` field is also required. It is used to determine how to retrieve the file.
+
+The `study_id` and `file_id` fields are allowed but are not required.
+
+
```
mfshao, line 38: Might be good to provide some code snippet in here as an example of how to use this `download_files_from_metadata()` in togather with external file retriever functions. Something similar to what you have in the example code provided in the team channel

gen3/tools/download/external_file_download.py
```diff
@@ -0,0 +1,256 @@
+"""
+Module for downloading and listing data from external repositories.
+
+    Examples:
+
+        See docs/howto/drsDownloading.md for more details
```
mfshao, line 6:
```suggestion
        See docs/howto/externalFileDownloading.md for more details
```

mfshao (approved):
awesome!

PR #225, opened 2024-06-17
Fix/drs pull hostname

thanh-nguyen-dang:
https://ctds-planx.atlassian.net/browse/MIDRC-635

### New Features
in case --endpoint is provided, we use endpoint to get files instead of look into the index


gen3/tools/download/drs_download.py
```diff
@@ -661,18 +661,20 @@ def resolve_drs_hostname_from_id(
...
     Args:
         object_ids (List[Downloadable]): list of object to resolve
         resolved_drs_prefix_cache (dict): cache of resolved DRS prefixes
         mds_url (str): Gen3 metadata service to resolve DRS prefixes
-
+        hostname (str): Hostname to main Gen3 environment
     """
     for entry in object_ids:
+        if endpoint is not None and entry.hostname is None:
+            entry.hostname = endpoint
```
mfshao, line 677: Not sure if this has been considered or if it is enven in scope, but in the context of Gen3 ecosystem, the `hostname` of a manifest entry is not necessarily going to be the `endpoint` value.
In case of `drs-pull` over a file manifest, this can be addressed by adding a optional `commons_url` field to manifest entries. 
I wonder if we should add a separated argument to the `drs-pull` over a file GUID case

thanh-nguyen-dang: @mfshao , I am not so sure about that. But, as a ticket stated, @frankliuao think that if we specify --endpoint to the command. It should download file from endpoint. And endpoint is used to set to hostname attribute of Downloader class.

So, I think this is the way to handle that and I have tried best to not touching to other case except the one when entry.hostname is null. I am not sure adding `commons_url` parameter to the command sdk solve the problem since we still at some point decide when we want to set value from that `commons_url` to the `entry.hostname` or when we use from `commons_url` and when we will try to resolve it by the referral sources.

paulineribeyre: @mfshao i don't think i understand the issue, could you provide some examples? or maybe we can huddle?

mfshao: Yeah sure, here is an example of how `drs-pull manifest` can work
With a file manifest of
```
[
  {
    "file_name": "test_file.zip",
    "file_size": 10,
    "object_id": "2275d149-0523-4428-bf37-fa4ccc0a0515",
    "commons_url": "qa-jcoin.planx-pla.net"
  }
]
```
And you can do this in qa-heal
```
gen3 --endpoint qa-heal.planx-pla.net drs-pull manifest manifest.json
```
In this case, the value of `endpoint` is not necessarliy the same as the `hostname` of the object GUID
I wonder if similarity feature should be added into the `drs-pull object` and `drs-pull objects` command along with this PR

paulineribeyre: ok i had a look at the code and i think this already works - if the manifest includes `commons_url` for an entry, then `entry.hostname` will not be None and we don't run the `if endpoint is not None and entry.hostname is None` block.

I think we need to make sure this is documented: `commons_url` takes precedence over `endpoint` and if neither is specified, we use the DRS prefix resolution.
(we should also test this in unit tests to make sure that's accurate)

tests/download_tests/test_download.py
```diff
@@ -625,7 +650,7 @@ def test_download_objects(
             )
             expected = {
                 "dg.XXTST/b96018c5-db06-4af8-a195-28e339ba815e": DownloadStatus(
-                    filename=None, status="error (resolving DRS host)"
+                    filename=None, status="error (no access methods)"
```
paulineribeyre, line 653: why has this changed?

tests/download_tests/test_download.py
```diff
@@ -600,6 +620,11 @@ def test_download_objects(
                 json={},
                 status_code=500,
             )
+            m.get(
+                "https://nullhost/ga4gh/drs/v1/objects/dg.XXTST/b96018c5-db06-4af8-a195-28e339ba815e",
+                json={},
+                status_code=500,
+            )
```
paulineribeyre, line 627: could you explain why you had to add this? since in this test we don't use the `endpoint` parameter i don't understand why this changed

gen3/tools/download/drs_download.py
mfshao: Yes this is already worked in `drs-pull manifest`, but I don't think it works for `drs-pull object` right? Like you cannot do `gen3 --endpoint qa-heal.planx-pla.net --commons_url qa-jcoin.planx-pla.net drs-pull object 2275d149-0523-4428-bf37-fa4ccc0a0515`, which is (arguable?) what part of this PR is for

paulineribeyre: oh i see, but in that case wouldn't you just do `gen3 --endpoint qa-jcoin.planx-pla.net drs-pull object 2275d149-0523-4428-bf37-fa4ccc0a0515` if you know if it's qa-jcoin?

mfshao: Yeah but that means you need to provide qa-jcoin's cred file.
In ecosystem, you can provide qa-heal's cred file, and the `drs-pull manifest` cmd will know it needs to talk to the qa-heal's MDS to get an access token for qa-jcoin to pull that file

tests/download_tests/test_download.py
thanh-nguyen-dang: that is to fix the test. The nullhost before cannot be resolve for DRS host since it does not exist in the list. But, but, because of the endpoint specified with endpoint and hostname. It will be resolved, but, it will have no access.

gen3/tools/download/drs_download.py
paulineribeyre: i see, that makes sense, i think it's not in scope for this ticket unless @thanh-nguyen-dang has time to add it, but maybe it can be noted as a todo when @thanh-nguyen-dang updates the doc

docs/howto/drsDownloading.md
```diff
@@ -73,6 +73,8 @@ To download a list of files, you can use the `drs-pull objects` command:
 gen3 --endpoint my-commons.org --auth <path to API key> drs-pull objects dg.XXTS/181af989-5d66-4139-91e7-69f4570ccd41 dg.XX22/221af922-2222-2239-22e7-62f4570cc222
 ```
 
+Note: if specified, the `commons_url` in the manifest file takes precedence over the `--endpoint` parameter. If neither is specified, DRS prefix resolution is used.
```
paulineribeyre, line 76: @thanh-nguyen-dang I updated the note ^

Did you test the specific use case in the ticket? (using a guid that has the MIDRC Prod prefix but using the endpoint param to download in MIDRC Staging or Validate)

frankliuao: ⬆️ Good point we should test it to make sure it works. 😄 eager to see the effect later when I need to pull data from validate.

thanh-nguyen-dang: @paulineribeyre , yes, I tested it

paulineribeyre (dismissed):

paulineribeyre (approved):

paulineribeyre (approved):

PR #228, opened 2024-06-25
Fix Gen3Index create_record to return the full record

k-burt-uch:
### Bug Fixes
- Fix Gen3Index create_record to return the created indexd record, rather than just the did, rev and base_id.


gen3/index.py
```diff
@@ -443,15 +442,9 @@ def create_record(
         }
         if did:
             json["did"] = did
-        resp = self.client._post(
-            "index/",
-            headers={"content-type": "application/json"},
-            data=client.json_dumps(json),
-            auth=self.client.auth,
-        )
-        raise_for_status_and_print_error(resp)
```
AlbertSnows, line 452: Do we care about preserving the print status? I didn't see it in the create function.

AlbertSnows (approved):
Approved assuming the answer is no to the print status question. Couldn't really read the html stuff but I assume that is auto generated.

gen3/index.py
k-burt-uch: I don't think so - it wasn't there prior to the change that caused the bug and indexclient handles the error here: https://github.com/uc-cdis/indexclient/blob/master/indexclient/client.py#L417 so its most likely redundant.

AlbertSnows (approved):

PR #234, opened 2024-09-13
Print errors in `submit_record`

paulineribeyre:
`raise_for_status_and_print_error(output)` was changed to `output.raise_for_status()` in https://github.com/uc-cdis/gen3sdk-python/pull/88. Reverting so that API errors are printed.

### Improvements
- `Gen3Submission.submit_record` now prints API errors


george42-ctds (approved):
Nice catch!

PR #244, opened 2024-11-23
Add Gen3 Workflow cli and tool.

nss10:
Link to JIRA ticket if there is one: [MIDRC-892](https://ctds-planx.atlassian.net/browse/MIDRC-892)

### New Features
* Adding Gen3 Wrapper command `gen3 run` that relays any command sent to it by adding `GEN3_TOKEN` env var. 

[MIDRC-892]: https://ctds-planx.atlassian.net/browse/MIDRC-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


paulineribeyre (commented):
Overall looks good, but i'm hoping we can find a way to forward _all_ the command arguments to nextflow. Meaning that whatever text is after `gen3 workflow` would be forwarded to nextflow directly. Not all nextflow commands start with "run", we shouldn't hardcode it.

For example: `gen3 workflow -C my.config run workflow.nf` should call `nextflow -C my.config run workflow.nf`

As an added bonus, this would make this tool generic - you could run `gen3 workflow --workflow-agent python script.py` and it would run `python script.py`.

If you get this to work, we could rename the tool so it's truly generic: `gen3 wrap <command> <arguments>` to run the command with a `GEN3_TOKEN` env var 🎉 

Please make sure to include documentation on the generic usage (if relevant) and the nextflow usage (such as using `$GEN3_TOKEN` in the config)

nss10:
If we were to achieve the gen3 wrap functionality, in order to run a nextflow workflow say on "main.nf", do we do, 

`gen3 wrap workflow run main.nf`

paulineribeyre:
@nss10 In that case i think there shouldn't be a default

gen3/tools/wrap.py
```diff
@@ -0,0 +1,32 @@
...
+        self.command_args = command_args
+
+    def run_command(self):
+        """
+        Take the command args and run a subprocess with appropriate access token in the env var
+        """
+        cmd = list(self.command_args)
+        try:
+            os.environ["GEN3_TOKEN"] = self.auth.get_access_token()
+            subprocess.run(cmd, stderr=subprocess.STDOUT)
```
paulineribeyre, line 30: could you investigate to see if unit tests can be written for this? i'm not sure what could be tested exactly, but if we can cover some of the code it would be nice.

gen3/tools/wrap.py
```diff
@@ -0,0 +1,32 @@
...
+        except Exception as e:
+            print("ERROR running command: ", e)
```
paulineribeyre, line 32:
```suggestion
            print(f"ERROR while running '{cmd}':", e)
```

gen3/tools/wrap.py
```diff
@@ -0,0 +1,32 @@
+# WIP: Explanation of the code, will be taken off once reviewed and approved
```
paulineribeyre, line 1: reminder to remove this comment and add docs

gen3/cli/wrap.py
```diff
@@ -0,0 +1,23 @@
...
+
+from gen3.tools.wrap import Gen3Wrap
+
+
+logger = get_logger("__name__")
+
+
+@click.command(
+    context_settings={"ignore_unknown_options": True, "allow_extra_args": True},
+    help="A wrapper command that forwards COMMAND_ARGS as-is with GEN3_TOKEN set to Gen3Auth's access token",
```
paulineribeyre, line 13:
```suggestion
    help="A wrapper command that forwards COMMAND_ARGS as-is after setting the environment variable GEN3_TOKEN",
```

gen3/cli/wrap.py
```diff
@@ -0,0 +1,23 @@
...
+)
+@click.argument("command_args", nargs=-1, type=click.UNPROCESSED)
+@click.pass_context
+def run(ctx, command_args):
+    auth = ctx.obj["auth_factory"].get()
+    gen3Wrap_object = Gen3Wrap(auth, command_args)
+    logger.info(
+        f"Running the command {command_args} with gen3 access token in environment variable"
```
paulineribeyre, line 21: I would remove this log and add a `f"Running '{cmd}'"` log to `run_command()` right before the command is run. This way if there is an error during the Gen3Auth logic, there won't be confusion about whether the error comes from the specified command

gen3/tools/wrap.py
nss10: I've added two unit tests

tests/test_wrap.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from unittest.mock import MagicMock, patch
+
+from gen3.tools.wrap import Gen3Wrap, Gen3Auth, Gen3AuthError
+
+
+# Patch subprocess.run and ensure appropriate args are sent to the method
```
paulineribeyre, line 7: could you move the description of the tests to docstrings instead of comments

tests/test_wrap.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+# Try to break the auth and see if Gen3AuthError is thrown
+@patch.object(Gen3Auth, "get_access_token", MagicMock(side_effect=Gen3AuthError()))
+def test_gen3_wrap_inavalid_auth(mock_gen3_auth):
+    test_command_args = ("echo", "Test1", "Test2")
+    with pytest.raises(Gen3AuthError):
+        with patch("gen3.tools.wrap.subprocess.run") as _:
+            wrapper_obj = Gen3Wrap(mock_gen3_auth, test_command_args)
+            wrapper_obj.run_command()
```
paulineribeyre, line 24: maybe you could also check that `mock_subprocess_run` was not called?

paulineribeyre (commented):
Looks great, minor comments

paulineribeyre (approved):

PR #246, opened 2024-12-17
Adding post_indexing_validation module

jacob50231:

### New Features
Adds a module for post-indexing validation


gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+            gen3file (object): An object that provides the `get_presigned_url` method to generate a pre-signed URL for the record.
+
+        Returns:
+            None: This function does not return any value. It modifies the `response_status` and `download_status` attributes.
+        """
+        logger.info(f"Checking record {self.guid}")
+        try:
+            resp = gen3file.get_presigned_url(self.guid)
+            url = resp.get("url")
+            response_status = 200
```
BedfordWest, line 64: This is a little strange to me - is there a reason we're manually setting the status code instead of logging what we actually got?

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+        download_success = -1
+        if response_status == 200:
+            try:
+                download_success = requests.get(url).status_code
+                logger.info(
+                    f"Download process complete with status code {download_success}"
+                )
+            except:
+                download_success = -1
+        self.download_status = download_success
```
BedfordWest, line 82: Shouldn't this line be inside the if block? I imagine we wouldn't want to set download_status to download_success if we weren't inside that block, right?

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+        return
+
+
+class Records:
```
BedfordWest, line 86: Having a "Record" class and "Records" class feels a bit like an anti-pattern to me. It might just be a naming issue, though. Docstrings for the classes explaining their intent might be helpful. If this class is meant more as a utility class to parse input into multiple records (which I think is the case after scanning over it a bit more), a more appropriate name/noun might be something like "RecordCreator" or "RecordParser".

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+    def read_records_from_manifest(self, manifest):
+        """
+        Parses a manifest and creates a dictionary of Record objects.
+
+        Args:
+            manifest (str): the location of a manifest file
+        """
+        if manifest[-3:] == "tsv":
```
BedfordWest, line 105: Might want to validate the length of the manifest before doing this just in case it's empty or some such and then throw a more helpful error rather than the default invalid index you'd get.

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+                )
+
+            for row in rows:
+                url_parsed = False
+                size = row["size"]
+                guid = row[guid_col]
+                for acl in row["acl"].split(" "):
+                    if acl != "admin":
+                        for url in row["url"].split(" "):
+                            if "://" not in url:
```
BedfordWest, line 128: Is there a reason we continue here rather than throwing an error? This would be an invalid url format, no?

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+                                continue
+                            else:
+                                protocol, bucket = (
+                                    url.split("://")[0].replace("[", ""),
+                                    url.split("/")[2],
+                                )
+                                key = (bucket, protocol, acl)
+                                if key not in self.record_dict or (
```
BedfordWest, line 136: The logic is getting a little funky here. It may be accurate, but a comment or docstring for the method might be useful to make it a little clearer. Possibly even worth splitting this out a little given how deeply nested it is. Maybe worth separating out url parsing for this purpose into its own utility? Just some thoughts.

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+                                        acl,
+                                        self.access_token,
+                                        self.commons,
+                                        size,
+                                    )
+                                    self.record_dict[key] = record
+                                url_parsed = True
+
+                if url_parsed == False:
+                    logger.warning(f"No url parsed for record {guid}")
```
BedfordWest, line 153: This feels like it might be more appropriate as an error, no? If we want to ensure everything gets validated and we don't exit early, we could store a list of anything that errored out and then throw an error at the end and provide the list.

gen3/tools/indexing/post_indexing_validation.py
```diff
@@ -0,0 +1,250 @@
...
+                    "acl": record.acl,
+                    "bucket": record.bucket,
+                    "protocol": record.protocol,
+                    "presigned_url_status": record.response_status,
+                    "download_status": record.download_status,
+                    "guid": record.guid,
+                }
+            )
+
+        self.download_results = download_results
```
BedfordWest, line 195: is there a reason we don't just use self.download_results from the start?

BedfordWest (changes requested):
A few thoughts/requests. I don't necessarily think all of these need to be addressed - we can maybe chat about it when we have a free cycle or eng workshop.

gen3/tools/indexing/post_indexing_validation.py
jacob50231: So instead of returning a requests.Response object Gen3File.get_presigned_url returns the json from that file which does not contain the status code. I'm only able to obtain a status code directly when raise_for_status_and_print_error raises an HTTPError from within get_presigned_url, but since that raises an error any time the status is 200, if it doesn't raise one I know the status is 200 so I set it manually. (I know it feels hacky and I hated doing it this way but it was the only way to do it without also updating the get_presigned_url function which would almost certainly cause breaking changes somewhere)

gen3/tools/indexing/post_indexing_validation.py
jacob50231: This was a bad name for this variable. I'm updating the script to not use the variable at all and instead directly change self.download_status

gen3/tools/indexing/post_indexing_validation.py
jacob50231: Yeah it's the latter. Renaming the class to make this more clear.

gen3/tools/indexing/post_indexing_validation.py
jacob50231: This should've been changed earlier when I added download_results as a class attribute. Good catch!

gen3/tools/indexing/post_indexing_validation.py
jacob50231: Will update the validation more here good catch!

gen3/tools/indexing/post_indexing_validation.py
jacob50231: Changed to throw an error

gen3/tools/indexing/post_indexing_validation.py
jacob50231: Set some comments to make it clearer. If you still want me to separate this out I can.

gen3/tools/indexing/post_indexing_validation.py
jacob50231: Updated

gen3/tools/indexing/post_indexing_validation.py
BedfordWest: Hmmmm. I think this is OK for now, but not ideal. For future reference, I'd probably just generalize the variable to something like "status_ok" as a bool and then in the log state we got a successful response. It might be the case that this always returns 200 for success now, but there are a bunch of types of successful HTTP responses and it just feels a bit too manual to me! 

But I think this is OK for now and no need to change. Thanks!

PR #249, opened 2025-02-10
Fix/external file retriever

mfshao:

### Improvements
- external file retrivers no longer requires `external_oidc_idp` as mandatory fields


piotrsenkow (approved):
Everything looks good to me

PR #254, opened 2025-04-25
PPS-2008 Fix/drs pull

mfshao:

### Breaking Changes
- If you intentionally what to skip using dataguids.org to resolve GUIDs for `drs-pull` command, you need to set the `--commons_url` option explicltly
### Bug Fixes
- Fixed a bug causing `drs-pull` command to skip DRS resolution when `--endpoint` or `GEN3_ENDPOINT` is set


gen3/tools/download/drs_download.py
```diff
@@ -800,6 +796,18 @@ def list_auth(hostname: str, authz: dict):
         print("      No access")
 
 
+def get_hostname_from_endpoint(endpoint: str):
```
mfshao, line 799: This is a temporary fix, since we are using the `endpoint` value differently across different packages
The `endpoint` value from `Gen3Auth` has protocol + hostname + port, whereas in `DRS download` what we actually need is only hostname part
We should consolidate the way we use `endpoint` across different places in Gen3SDK, but I think that should be deferred to another ticket

Avantol13 (approved):

PR #268, opened 2025-07-18
Feat : Multiple download functionality

Dhiren-Mhatre:
New Features

Dependency updates

### DIFF
--- a/gen3/file.py
+++ b/gen3/file.py
@@ -27,7 +27,6 @@ MAX_RETRIES = 3
 DEFAULT_NUM_PARALLEL = 3
 DEFAULT_MAX_CONCURRENT_REQUESTS = 50
 DEFAULT_QUEUE_SIZE = 1000
-DEFAULT_BATCH_SIZE = 100
 
 
 class Gen3File:
@@ -396,7 +395,6 @@ class Gen3File:
         max_concurrent_requests=DEFAULT_MAX_CONCURRENT_REQUESTS,
         num_processes=DEFAULT_NUM_PARALLEL,
         queue_size=DEFAULT_QUEUE_SIZE,
-        batch_size=DEFAULT_BATCH_SIZE,
         skip_completed=False,
         rename=False,
         no_progress=False,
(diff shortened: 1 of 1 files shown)


Avantol13 (changes requested):
Great start, very rough initial review comments. Please make sure to update your PR description and remove any bullets for categories where there's nothing (such as breaking changes). For other categories, please try to summarize the feature/updates in 1-2 bullets. If you want the other information to be there, make sure it's placed *above* the markdown headers. This is all b/c we automatically parse the PR description to generate release notes and we only want a bullet or two in each header for a single PR. It should be concise and clear without knowing what PR it came from

gen3/file.py
```diff
@@ -259,3 +374,402 @@ def upload_file_to_guid(
...
+                        async for chunk in resp.content.iter_chunked(chunk_size):
+                            await f.write(chunk)
+                    return True
+                return False
+        except Exception as e:
+            logging.warning(f"Download failed for {guid}: {e}")
+            return False
+
+    @staticmethod
+    def _format_filename_static(guid, original_filename, filename_format):
```
paulineribeyre, line 747: why do we need both `_format_filename` and `_format_filename_static`, and why is the logic different?

Dhiren-Mhatre:
I've addressed all the feedback from the code review and updated the implementation accordingly. Thanks!

gen3/file.py
```diff
@@ -259,3 +385,434 @@ def upload_file_to_guid(
...
+                    batch_results.append(result)
+            except Exception as e:
+                logging.error(f"Error processing {guid} in batch: {e}")
+                batch_results.append(
+                    {"guid": guid, "status": "failed", "error": str(e)}
+                )
+        return batch_results
+
+    @staticmethod
+    async def _download_single_guid(
```
paulineribeyre, line 634: It's a bit confusing to see similarly named functions: `_download_single_guid` and `download_single`.
Also `_download_file_worker` which is only used for download-single and `_worker_process` which is only used for download-multiple.
I'm wondering if the download-single logic could just reuse some of the download-multiple logic?

Dhiren-Mhatre:
@paulineribeyre  All of the above comments and issues have been resolved in the newly created pull request, which includes both multiple downloading functionality and performance testing. We can close this pull request and the previous pull request.

PR #271, opened 2025-08-05
Feat : Multiple download functionality  

Dhiren-Mhatre:
New Features
* Performance Testing Suite: Added comprehensive benchmarking framework comparing the new implementation against CDIS Data Client for download speed, memory usage, and concurrency efficiency

Dependency updates


Avantol13 (changes requested):
I haven't had time to review other Python code, will get to that soon

docs/howto/asyncDownloadMultiple.md
```diff
@@ -0,0 +1,256 @@
...
+    skip_completed=True,
+    no_progress=False
+))
+
+print(f"Succeeded: {len(result['succeeded'])}")
+print(f"Failed: {len(result['failed'])}")
+print(f"Skipped: {len(result['skipped'])}")
+```
+
+## Parameters
```
Avantol13, line 88: let's not include this in the markdown b/c the code may change, instead: make sure the information here exists in the command line help arguments directly when you run `poetry run gen3 download-multiple-async --help` but remove it from the markdown. Adding the code docs here creates a risk of drift and things getting out of sync.

gen3/cli/download.py
```diff
@@ -0,0 +1,198 @@
...
+@click.option(
+    "--filename-format",
+    default="original",
+    type=click.Choice(["original", "guid", "combined"]),
+    help="Filename format: 'original' uses the original filename from metadata, 'guid' uses only the file GUID, 'combined' uses original filename with GUID appended",
+)
+@click.option("--protocol", default=None, help="Protocol for presigned URL (e.g., s3)")
+@click.option("--skip-completed", is_flag=True, help="Skip files that already exist")
+@click.option("--rename", is_flag=True, help="Rename file if it already exists")
+@click.option("--no-prompt", is_flag=True, help="Do not prompt for confirmations")
```
Avantol13, line 61: lets make sure we have defaults made clear in these

gen3/cli/download.py
```diff
@@ -0,0 +1,198 @@
...
+    "--filename-format",
+    default="original",
+    type=click.Choice(["original", "guid", "combined"]),
+    help="Filename format: 'original' uses the original filename from metadata, 'guid' uses only the file GUID, 'combined' uses original filename with GUID appended",
+)
+@click.option("--protocol", default=None, help="Protocol for presigned URLs (e.g., s3)")
+@click.option(
+    "--max-concurrent-requests",
+    default=10,
+    help="Maximum concurrent async downloads",
```
Avantol13, line 116: let's make sure defaults are clear in the --help

gen3/cli/download.py
```diff
@@ -0,0 +1,198 @@
...
+            return
+
+        if not no_prompt:
+            click.echo(f"Found {len(manifest_data)} files to download")
+            if not click.confirm("Continue with async download?"):
+                click.echo("Download cancelled")
+                return
+
+        file_client = Gen3File(auth_provider=auth)
+
```
Avantol13, line 158: let's add a debug log here that clearly states the inputs (manifest_data, download, filename format, protocol, max_concurrent_r., skip completed, rename, no_progress).

gen3/cli/download.py
```diff
@@ -0,0 +1,198 @@
...
+                skip_completed=skip_completed,
+                rename=rename,
+                no_progress=no_progress,
+            )
+        )
+
+        click.echo(f"\nAsync Download Results:")
+        click.echo(f"✓ Succeeded: {len(result['succeeded'])}")
+
+        if len(result["skipped"]) > 0:
```
Avantol13, line 176: Is it possible result["skipped"] is a string? B/c I'm still seeing the full output:

Async Download Results:
✓ Succeeded: 2071
- Skipped: 0
✗ Failed: 0

I don't want to see skipped or failed if it's zero

gen3/cli/download.py
```diff
@@ -0,0 +1,198 @@
...
+    try:
+        loop = asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+    return loop
+
+
+def load_manifest(manifest_path: str) -> List[Dict[str, Any]]:
+    """Load manifest from JSON file."""
```
Avantol13, line 27: make sure docstrings have arguments in Google style, see other places in the SDK for examples.

Avantol13 (changes requested):

gen3/cli/download.py
```diff
@@ -0,0 +1,275 @@
...
+            return False
+
+    return True
+
+
+@click.command()
+@click.argument("guid")
+@click.option(
+    "--download-path",
+    default=".",
```
Avantol13, line 71: let's make this default more dynamic and a folder by default.

maybe let's do a timestamped folder name like: 

```
from datetime import datetime

...

default=f"download_{datetime.now().strftime("%d_%b_%Y")}",

```

Avantol13: do the same thing for multiple download async

gen3/cli/download.py
```diff
@@ -0,0 +1,275 @@
...
+    help="Filename format: 'original' uses the original filename from metadata, 'guid' uses only the file GUID, 'combined' uses original filename with GUID appended (default: original)",
+)
+@click.option(
+    "--protocol",
+    default=None,
+    help="Protocol for presigned URLs (e.g., s3) (default: auto-detect)",
+)
+@click.option(
+    "--max-concurrent-requests",
+    default=300,
```
Avantol13, line 160: this is really, really high. Let's scale this default to something like 20 and people can manually bump this up as needed.

gen3/cli/download.py
```diff
@@ -0,0 +1,275 @@
...
+        if not no_prompt:
+            click.echo(f"Found {len(manifest_data)} files to download")
+            if not click.confirm("Continue with async download?"):
+                click.echo("Download cancelled")
+                return
+
+        file_client = Gen3File(auth_provider=auth)
+
+        # Debug logging for input parameters
+        logging.debug(
```
Avantol13, line 230: put this before the prompt above and you need to use our logger. Put this at the top of the file. See other files for reference:


```
from cdislogging import get_logger

logging = get_logger("__name__")
```

gen3/cli/download.py
```diff
@@ -0,0 +1,275 @@
...
+    type=int,
+)
+@click.option(
+    "--queue-size",
+    default=1000,
+    help="Maximum items in input queue (default: 1000)",
+    type=int,
+)
+@click.option(
+    "--skip-completed",
```
Avantol13, line 177: I'm not sure this is working as intended. I get a lot of warnings when I try this:

```
$ poetry run gen3 -vv download-multiple-async --manifest MIDRC_case_manifest.json --download-path ./downloads --skip-completed 
[2025-09-02 14:28:36,954][  DEBUG] Initializing auth..
Found 5178 files to download
Continue with async download? [y/N]: y
Downloading:   0%|                                                                                                                                                                                                                      | 0/5178 [00:00<?, ?it/s][2025-09-02 14:28:40,004][WARNING] File will be overwritten: downloads/10041569-u_H3HaB1lES6-HXVpiEfMA/2.16.840.1.114274.1818.48858790339993589885669552017090272896/2.16.840.1.114274.1818.56909342758958044433235758152671341713.zip
[2025-09-02 14:28:40,010][WARNING] File will be overwritten: downloads/10041569-FtpnR4GEPEe9JztuDUqrXg/2.16.840.1.114274.1818.544490635779373958610541144202466729913/2.16.840.1.114274.1818.55741368199706394947919032241968638388.zip
[2025-09-02 14:28:40,085][WARNING] File will be overwritten: downloads/10041569-CooqJZVIRkywB6_o9CJi0Q/2.16.840.1.114274.1818.567640554746210778914433462824885428113/2.16.840.1.114274.1818.523353596424499437510054502719882590104.zip

```

Actually... it seems like maybe it is actually skipping but we need to suppress all these warnings and/or not overwrite existing files, just skip downloading again.

gen3/cli/download.py
```diff
@@ -0,0 +1,275 @@
...
+    is_flag=True,
```
Avantol13, line 178: I don't think you can have an is_flag and default to True b/c how do I set this to false then? it just needs to be a boolean but default to true

Avantol13: you could reverse the logic here and just call it: `redownload-completed` and then leave is_flag=True but default to false

Avantol13: make sure to update docs if you change this

docs/howto/asyncDownloadMultiple.md
```diff
@@ -0,0 +1,198 @@
...
+### Configuration Recommendations
+
+For optimal performance, adjust the concurrency and process settings based on your specific use case:
+
+- **Small files**: Use higher concurrent request limits
+- **Large files**: Use lower concurrent request limits to avoid overwhelming the system
+- **High-bandwidth networks**: Increase the number of worker processes
+- **Limited memory**: Reduce queue sizes to manage memory usage
+
+### Memory Management
```
Avantol13, line 114: let's remove this section, I feel it's redundant given the above config info

docs/howto/asyncDownloadMultiple.md
```diff
@@ -0,0 +1,198 @@
...
+
+- **Queue Size**: Adjust based on available system memory
+- **Batch Size**: Balance between memory usage and processing overhead
+- **Process Count**: Match available CPU cores for optimal performance
+
+### Network Optimization
```
Avantol13, line 120: let's also remove this section

docs/howto/asyncDownloadMultiple.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Troubleshooting
+
+### Common Issues
+
+**Slow Downloads:**
+
+- Check network bandwidth and server limits
+- Reduce concurrent request limits if server is overwhelmed
+- Verify authentication token is valid
```
Avantol13, line 145: auth token would not be a reason for slow downloads, let's remove this

Avantol13 (changes requested):

gen3/cli/download.py
Avantol13: don't change it, actually. We need to ensure we're **matching** previous commands exactly.

Here's the help from gen3-client:

```
Flags:
      --download-path string     The directory in which to store the downloaded files (default ".")
      --filename-format string   The format of filename to be used, including "original", "guid" and "combined" (default "original")
  -h, --help                     help for download-multiple
      --manifest string          The manifest file to read from. A valid manifest can be acquired by using the "Download Manifest" button in Data Explorer from a data common's portal
      --no-prompt                If set to true, will not display user prompt message for confirmation
      --numparallel int          Number of downloads to run in parallel (default 1)
      --protocol string          Specify the preferred protocol with --protocol=s3
      --rename                   Only useful when "--filename-format=original", will rename file by appending a counter value to its filename if set to true, otherwise the same filename will be used
      --skip-completed           If set to true, will check for filename and size before download and skip any files in "download-path" that matches both
```

Let's change the overall command name to match too.   `download-multiple` instead of `download-multiple-async`

Keep skip completed but make it a boolean so I can pass in false.

I'm okay with --numparallel not existing and being different b/c the parallelization scheme in Python is a bit different... maybe we can allow that to be passed in though and just use the value as `max-concurrent-requests` if that isn't supplied. So people can easily port their commands over and have them work right away

Avantol13 (changes requested):
Make sure to add unit tests for this new functionality

tests/test_file.py
```diff
@@ -395,3 +396,113 @@ def test_upload_file_wrong_api_key(gen3_file, supported_protocol, authz, expires
...
+@pytest.fixture
+def mock_manifest_data():
+    return [
+        {"guid": "test-guid-1", "file_name": "file1.txt"},
+        {"guid": "test-guid-2", "file_name": "file2.txt"},
+        {"object_id": "test-guid-3", "file_name": "file3.txt"},
+    ]
+
+
+def test_download_single_success(gen3_file):
```
Avantol13, line 410: please add a docstring for each test describing what you're testing

docs/howto/asyncDownloadMultiple.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### High-Performance Configuration
+
+```bash
+# Optimized for high-throughput downloads
+gen3 --endpoint data.commons.io --auth creds.json download-multiple \
+    --manifest large_dataset.json \
+    --download-path ./large_downloads \
+    --max-concurrent-requests 20 \
```
Avantol13, line 181: let's change to:
```suggestion
    --max-concurrent-requests 64 \
    --max-concurrent-requests 8 
```

Avantol13 (dismissed):
You will also need to update this test. I got it working:

```python

def test_download_single_basic_functionality(gen3_file):
    """
    Test download_single basic functionality with synchronous download.

    Verifies that download_single downloads a file successfully using
    synchronous requests and returns True.
    """
    gen3_file._auth_provider._refresh_token = {"api_key": "123"}

    with patch.object(gen3_file, 'get_presigned_url') as mock_presigned, \
         patch('gen3.file.requests.get') as mock_get, \
         patch('gen3.index.Gen3Index.get_record') as mock_index:

        mock_presigned.return_value = {"url": "https://fake-url.com/file"}
        mock_index.return_value = {"file_name": "test-file.txt"}
        mock_response = MagicMock()
        mock_response.status_code = 200
        mock_response.headers = {"content-length": "12"}
        mock_response.iter_content = lambda size: [b"test content"]
        mock_get.return_value = mock_response

        result = gen3_file.download_single(object_id="test-guid", path="/tmp", protocol="s3")

        assert result == True
        mock_presigned.assert_called_once_with("test-guid", protocol="s3")
        mock_index.assert_called_once_with("test-guid")
```

jacob50231 (dismissed):

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+@click.option(
+    "--max-concurrent-requests",
+    default=20,
+    help="Maximum concurrent async downloads per process (default: 20)",
+    type=int,
+)
+@click.option(
+    "--numparallel",
+    default=None,
+    help="Number of downloads to run in parallel (compatibility with gen3-client)",
```
Avantol13:
```suggestion
    help="Same as max-concurrent-requests, for compatibility with previous gen3-client arguments",
```

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+        if result["failed"] and len(result["failed"]) > 0:
+            click.echo(f"✗ Failed: {len(result['failed'])}")
+
+        if result["failed"]:
+            click.echo("\nFailed downloads:")
+            for failure in result["failed"]:
+                click.echo(
+                    f"  - {failure.get('guid', 'unknown')}: {failure.get('error', 'Unknown error')}"
+                )
+
```
Avantol13: this is not working, I'm getting this:

```
[2026-01-21 15:19:03,719][  ERROR] Async batch download failed: 'str' object has no attribute 'get'
[2026-01-21 15:19:03,719][  ERROR] Async batch download failed: 'str' object has no attribute 'get'
```

if I provide the command `poetry run gen3 download-multiple --manifest MIDRC_case_manifest_small.json --no-progress --no-prompt --protocol foo --skip-completed false`

I suggest we remove this listing of failures entirely, as the ERROR logs themselves have the GUIDs that failed in them

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+            click.echo(
+                "\nTo retry failed downloads, run the same command with --skip-completed flag:"
```
Avantol13:
```suggestion
                "\nTo retry failed downloads, run the same command with --skip-completed true"
```

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+
+        success_rate = len(result["succeeded"]) / len(manifest_data) * 100
```
Avantol13: we should rethink this because the way it's working, I get a non 100 success if I skip previously downloaded files. 

```
$ poetry run gen3 -vv download-multiple --manifest MIDRC_case_manifest.json --no-prompt --max-concurrent-requests 64 --num-processes 8

...

Async Download Results:
✓ Succeeded: 4871
- Skipped: 307

Success rate: 94.1%
```

If --skip-completed is true, we should count skipped towards success imo

Avantol13: this is even more apparent when I rerun a successful download and it skips everything

```
Async Download Results:
✓ Succeeded: 0
- Skipped: 42

Success rate: 0.0%
```

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+    no_prompt=False,
+    no_progress=False,
+):
+    """
+    Asynchronously download multiple files from a manifest with just-in-time presigned URL generation.
+    """
+    auth = ctx.obj["auth_factory"].get()
+
+    # Use numparallel as max_concurrent_requests if provided (for gen3-client compatibility)
+    if numparallel is not None and max_concurrent_requests == 20:  # 20 is the default
```
Avantol13: create a global for this default value, don't hard code and make a magic number

gen3/cli/download.py
```diff
@@ -0,0 +1,265 @@
...
+    default=f"./download_{datetime.now().strftime('%d_%b_%Y')}",
+    help="Directory to download file to (default: timestamped folder)",
+)
+@click.option(
+    "--protocol",
+    default=None,
+    help="Protocol for presigned URL (e.g., s3) (default: auto-detect)",
+)
+@click.pass_context
+def download_single(
```
Avantol13, line 107: I would like this support the same interface as the gen3-client, and ideally the same params (the ones that translate) from download-multiple. What we could maybe do is just wrap download-multiple behind the scenes but pass a manifest with a single GUID (e.g. don't repeat code we don't need to repeat).

As a reference, this is gen3-client's support:

```cli
$ gen3-client download-single --help                                                                        
Gets a presigned URL for a file from a GUID and then downloads the specified file.

Usage:
  gen3-client download-single [flags]

Examples:
./gen3-client download-single --profile=<profile-name> --guid=206dfaa6-bcf1-4bc9-b2d0-77179f0f48fc

Flags:
      --download-path string     The directory in which to store the downloaded files (default ".")
      --filename-format string   The format of filename to be used, including "original", "guid" and "combined" (default "original")
      --guid string              Specify the guid for the data you would like to work with
  -h, --help                     help for download-single
      --no-prompt                If set to true, will not display user prompt message for confirmation
      --protocol string          Specify the preferred protocol with --protocol=gs
      --rename                   Only useful when "--filename-format=original", will rename file by appending a counter value to its filename if set to true, otherwise the same filename will be used
      --skip-completed           If set to true, will check for filename and size before download and skip any files in "download-path" that matches both

Global Flags:
      --profile string   Specify profile to use
```

Avantol13 (changes requested):

Avantol13 (changes requested):
in general things are looking good but see my comments. It's probably prudent to test using Python 3.13 (I did locally and had to fix a dep issue, I think just a bump, but double check that).

gen3/cli/download.py
```diff
@@ -116,6 +118,93 @@ def download_single(
...
+    is_flag=True,
+    help="Rename file if it already exists (default: false)",
+)
+@click.option(
+    "--no-progress",
+    is_flag=True,
+    help="Disable progress bar (default: false)",
+)
+@click.pass_context
+def download_single(
```
Avantol13, line 107: this was already defined and is now being overridden. remove the old code. 

Also can you adjust so that the GUID is not an arg but a --guid VALUE to match the gen3-client exactly?

Avantol13: I'm not sure if this causes the issues I'm seeing, but download-single is hanging for me and not completing:

```
poetry run gen3 -vv d-vv download-single dg.MD1R/28fb1a28-d662-429b-9b74-91b2aa03c63f                                                           
[2026-02-12 12:34:35,203][  DEBUG] Initializing auth..
Downloading:   0%|                                                                                                         | 0/1 [00:00<?, ?it/s]

```

it just hangs there forever

Avantol13 (changes requested):

PR #272, opened 2025-09-08
feat:added profile functionality

Dhiren-Mhatre:
### New Features

  - Added profile-based authentication support compatible with cdis-data-client
  - New CLI commands: gen3 configure, gen3 list-profiles, gen3 show-profile
  - Global --profile option for all gen3 commands
  - Cross-compatibility with both ~/.gen3/config and ~/.gen3/gen3_client_config.ini formats


Avantol13 (dismissed):

jacob50231 (approved):

Avantol13:
I think you've got to update for 3.13,

change to 

```
for ep in entry_points().select(group="..."):
```

vs.

```
for ep in entry_points().get("...", [])
```

in pfb.py and users.py

at least I had to for it to work. Double check that also works in 3.9 and if there's a way to support both, let's do it

Avantol13 (changes requested):
I'd like to refactor some of the stuff here - the profile listing support it's polluting the top level commands.

Let's see if we can just move those under `configure`. e.g.

instead of top level commands, let's do this:

```
gen3 configure list-profiles
gen3 configure show-profile --profile MIDRC_prod
```

and then the existing:

`gen3 configure --profile new_profile ...`

PR #285, opened 2026-01-02
PD-97 Add unit test to check CLI/pypfb import

paulineribeyre:
Regression test for:
```
Traceback (most recent call last):
  File "/Users/kyle/.pyenv/versions/3.13.4/bin/gen3", line 5, in <module>
    from gen3.cli.__main__ import main
  File "/Users/kyle/.pyenv/versions/3.13.4/lib/python3.13/site-packages/gen3/cli/__main__.py", line 9, in <module>
    import gen3.cli.pfb as pfb
  File "/Users/kyle/.pyenv/versions/3.13.4/lib/python3.13/site-packages/gen3/cli/pfb.py", line 2, in <module>
    from pfb import cli as pfb_cli
  File "/Users/kyle/.pyenv/versions/3.13.4/lib/python3.13/site-packages/pfb/cli.py", line 63, in <module>
    for ep in entry_points().get("pfb.plugins", []):
              ^^^^^^^^^^^^^^^^^^
```

Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PD-97

### Bug Fixes
- Fixes issue with SDK running in python 3.9


tests/test_cli.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+def test_import_cli(monkeypatch):
```
Avantol13, line 4:
```suggestion
def test_import_cli(monkeypatch):
     """
     Test that invoking the help from the CLI does not return an error
     """
```

gen3/utils.py
```diff
@@ -341,6 +341,23 @@ def deep_dict_update(a, b):
...
+def load_entry_points():
+    try:
+        from importlib.metadata import entry_points
+    except ImportError:
+        from importlib_metadata import entry_points
+
+    # Load plug-ins from entry_points (syntax changes for python 3.12+)
+    major = sys.version_info[0]
+    minor = sys.version_info[1]
+    if major == 3 and minor >= 12:
```
Avantol13, line 353: I would much rather try/catch than explicitly name major and minor versions in the code, but I don't feel super strongly about it. It's just that if in the long long future Python releases a 4 major, this is going to break again.

Avantol13 (dismissed):

gen3/utils.py
```diff
@@ -341,6 +341,23 @@ def deep_dict_update(a, b):
...
+        for ep in entry_points(group="pfb.plugins"):
+            ep.load()
+    else:
+        for ep in entry_points().get("pfb.plugins", []):
+            ep.load()
```
paulineribeyre: i didn't look into how this works exactly, but it looks off to me that it's hardcoded to `pfb.plugins` here and called in both `gen3/cli/pfb.py` and `gen3/cli/users.py`

k-burt-uch: Fixed. It should be gen3.plugins not pfb.plugins. Thanks for catching the copy paste error

paulineribeyre:
See https://cdis.slack.com/archives/C0294A7KVE0/p1772573412435139?thread_ts=1772558196.769849&cid=C0294A7KVE0

Issue #222, opened 2024-06-06
BUG: get_dbgap_accession_as_parts does not correctly parse a full dbgap accession


SpencerAxelrod:
`gen3.external.nih.utils.get_dbgap_accession_as_parts` does not seem to parse a full dbgap accession.

Example: 

`get_dbgap_accession_as_parts("phs001961.v2.p1.c1")`

returns

```
{
'phsid': 'phs001961', 
'phsid_number': '001961', 
'participant_set': '', 
'participant_set_number': '',
'version': 'v2', 
'version_number': '2', 
'consent': '', 
'consent_number': ''
}
```
with participant set and consent group information missing

AkshataAthreya:
Hi! I’m Akshata — a CS undergrad working mainly with Python and data analysis. I recently interned at Siemens Gamesa (now Siemens Energy), where I worked on Python automation and unit testing.

I’m exploring the uc-cdis codebase as I plan to apply for Google Summer of Code 2026, and I wanted to ask if this issue would be suitable for a first-time contributor. The bug description and example are very clear, and I’d be happy to take a stab at fixing it and adding tests if that’s helpful.

Please let me know if there are any guidelines or context I should be aware of before starting. Thanks!

WilsonnnTan:
Hi, I'm Wilson 👋 
Recently, I'm exploring uc-cdis gen3 SDK codebase as my prep for GSoC 2026 and I would like to work on this problem.
Thanks!

WilsonnnTan:
Hi, I triaged this issue since there hasn’t been recent maintainer activity on this issue, I investigated it independently to clarify the bug and found that the root cause appears to be in the parser regex logic.

PR #290, opened 2026-02-09
fix(dbgap): dbgap accession parser

WilsonnnTan:
- Fix dbgap accession regex logic. Issue: #222


gen3/external/nih/utils.py
```diff
@@ -7,8 +7,8 @@
 # For more details about this regex, see the function that uses it
 DBGAP_ACCESSION_REGEX = (
     "(?P<phsid>phs(?P<phsid_number>[0-9]+))"
-    "(.(?P<participant_set>p(?P<participant_set_number>[0-9]+))){0,1}"
     "(.(?P<version>v(?P<version_number>[0-9]+))){0,1}"
+    "(.(?P<participant_set>p(?P<participant_set_number>[0-9]+))){0,1}"
```
k-burt-uch, line 11: This is a great catch, thank you. 

Could you write unit tests for `get_dbgap_accession_as_parts` that ensures these parts are still returned properly? 

You can create a new test_utils.py file that verifies that the `p` number maps to the participant_set and the `v` maps to the version_number returned by the function.

k-burt-uch (changes requested):

gen3/external/nih/utils.py
WilsonnnTan: Hi, thanks for the response. I've created the test cases for it

tests/test_utils.py
```diff
@@ -0,0 +1,66 @@
...
+            'version': 'v3',
+            'version_number': '3',
+            'participant_set': '',
+            'participant_set_number': '',
+            'consent': 'c3',
+            'consent_number': '3'
+        }
+    ),
+    (
+        'phs000123.v3.p1.c3',
```
k-burt-uch, line 47: Could you add a case where the v, p and c numbers are not in the expected order for the four number version and the three number version? 

Something like 
phs000123.p3.v1.c3
phs000123.v3.c1.p3

and 
phs000123.p3.v1
phs000123.v3.c1

k-burt-uch (changes requested):

tests/test_utils.py
WilsonnnTan: @k-burt-uch Are the utils expected to successfully parse it or not? I tested it here, and it failed to parse.

k-burt-uch: They're not expected to parse correctly, but I'm hoping to avoid a situation where they're parsed incorrectly and information is returned, such as 

`  phs000123.v2.c1.p3`
parsing to 

```
{
            'phsid': 'phs000123',
            'phsid_number': '000123',
            'version': 'v2',
            'version_number': '2',
            'participant_set': '1',
            'participant_set_number': '1',
            'consent': 'c3',
            'consent_number': '3'
        }
```

to avoid situations like this where the c and p numbers are switched.

k-burt-uch (approved):

PR #309, opened 2026-07-17
Chore/merge gsoc

jacob50231:
New Features
* Performance Testing Suite: Added comprehensive benchmarking framework comparing the new implementation against CDIS Data Client for download speed, memory usage, and concurrency efficiency

Dependency updates


gen3/file.py
```diff
@@ -259,3 +278,438 @@ def upload_file_to_guid(
...
+            enable_cleanup_closed=True,
+        )
+        semaphore = asyncio.Semaphore(max_concurrent)
+
+        async with aiohttp.ClientSession(
+            connector=connector, timeout=timeout
+        ) as session:
+            while True:
+                try:
+                    # Check if queue is empty with timeout
```
SpencerAxelrod, line 462: Is there supposed to be a timeout around here? I see a commit called "removed timeout" in the history. But also, should we not send a better completion signal here (instead of a timeout)?

SpencerAxelrod: like have `_guid_producer` put one `None` (`input_queue.put(None)`) per worker after the GUIDs. Have this loop break when it reads one.

gen3/file.py
```diff
@@ -259,3 +278,438 @@ def upload_file_to_guid(
...
+
+            results = {"succeeded": [], "failed": [], "skipped": []}
+            completed_count = 0
+
+            if not no_progress:
+                pbar = tqdm(total=len(guids), desc="Downloading")
+
+            while completed_count < len(guids):
+                try:
+                    batch_results = output_queue.get()
```
SpencerAxelrod, line 353: if every worker dies before producing all len(guids) results, this blocks forever instead of erroring out and the except Empty right below never gets hit

PR #313, opened 2026-08-19
Bulk FHIR Gen 3 Authorization Tagging 

dsafarian:
### New Features
 - Bulk FHIR Gen 3 Authorization Tagging (see more details with gen3 fhir --help or in the fhir.md in docs/howto)

### Dependency updates
 - fhirpathpy
 
### Breaking Changes
 - Drop support for Python 3.9, which has been end-of-life since last year and nedded dependencies no longer support it


pyproject.toml
```diff
@@ -53,6 +54,7 @@ requests-mock = "*"
 cdisutilstest = { git = "https://github.com/uc-cdis/cdisutils-test.git", tag = "2.0.0" }
 indexd = { git = "https://github.com/uc-cdis/indexd.git", tag = "5.0.4" }
 deptry = "^0.23.1"
+setuptools = "<82"
```
Avantol13: why was this necessary? let's try without it

docs/howto/fhir.md
```diff
@@ -0,0 +1,14 @@
...
+
+The fhir commands can be invoked as follows
+
+`gen3 fhir COMMAND [ARGS] [OPTIONS]`
+
+For a list of commands and options run
+
+`gen3 fhir --help`
+
+For example, the following tags the 'Patient.ndjson' file with Gen3 authorization and outputs 'gen3_Patient.ndjson' using the authorization rules from 'config.yaml'
```
Avantol13, line 16: can you add some more details about the configuration and how to configure different rules?

docs/howto/fhir.md
```diff
@@ -0,0 +1,14 @@
+## FHIR
+
```
Avantol13, line 7: We should have a few sentences background here. Expand what FHIR stands for. Say Gen3 is working on adding support and that this CLI will help with data preparation and potentially interaction in the future

gen3/cli/fhir.py
```diff
@@ -0,0 +1,107 @@
...
+def cleanup(dry_run: bool, force: bool):
+    """
+    Remove all intermediate files in the tmp folder from previous runs
+
+    Args:
+        dry_run (bool): If True, list the files that would be removed, but not actually remove them
+        force (bool): Delete all intermediate directories disregarding the status
+    """
+
+    sweep(dry_run=dry_run, force=force)
```
Avantol13: let's name this function cleanup_fhir_transform_artifacts or something more descriptive

tests/test_data/fhir_config.yaml
Avantol13, line 1: let's add a README.md to this test_data directory that explains that this is all synthetic data generated

Avantol13: and clarify which FHIR implementation guide / model. I assume this is the default FHIR model. Did you use a specific FHIR version?

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+BATCH_SIZE = 1
+BASE_RECORD = {
+    "timestamp": "2026-08-12T21:00:20.677168+00:00",
+    "input_file": "./tests/test_data/Patient.ndjson",
+    "output_file": "./tests/test_data/fhir_outputs/status.ndjson",
+    "config_hash": "9c61e70e3ea403e3daf1ee795036253c",
+    "batch_size": 1,
+}
+
+os.makedirs(TMP_ROOT, exist_ok=True)
```
Avantol13: don't make calls like this in global context - you can have a session-level fixture that guarantees this runs only once in pytest

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+import time
+from datetime import datetime, timezone
+import shutil
+import pathlib
+import hashlib
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
+from cdislogging import get_logger
+from gen3.utils import make_folders_for_filename
+
+logging = get_logger("fhir_transform", log_level="info")
```
Avantol13:
```suggestion
logging = get_logger(__name__)
```

you should just use the gen3.fhir name from the dunder python provides. Also, don't override the log_level, the top-level gen3 logger sets it

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+
+    """
+
+    def __init__(self, config_path: str | os.PathLike[str], custom_hook=None):
+        """
+        Initialize instance of the tagger.
+
+        Args:
+            config_path (str): the name/path of the config.yaml file
+            custom_hook (?): ???
```
Avantol13: custom_hook is a Callable

from collections.abc import Callable

Avantol13: the docs should describe where it's called. e.g. if provided, called instead of using configuration to determine authz. it passes the FHIR resource as a dict and expects the authz resource

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def determine_authz(self, resource: dict) -> str:
+        """
+        Determines the correct authorization given the resource type and condition
+
+        Args:
+            resource (dict): the resource dictionary (a line of the .ndjson file)
+
+        Returns:
+            The appropriate rule/authorization for the resource
```
Avantol13: You need to specify return type

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+                return hook_result
+
+        # check global catch-all fallback
+        if "global_authz" in self.config:
+            return self.config["global_authz"]
+
+        # resource_type = resource.get("resourceType")
+
+        # check static rules engine via FHIRPath
+        #    for rule in self.config.get("rules", []):
```
Avantol13: this is dead code, delete it

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+        #        if rule["resource_type"] == resource_type:
+        #            match = evaluate(resource, rule["condition"])  #how to process in bulk
+        #            if match and match[0] is True:
+        #                return rule["authz"]
+        for rule in self.rules:
+            match = evaluate(resource, rule["condition"])
+            if match and match[0] is True:  # FIXME: what if multiple conditions match?
```
Avantol13: for the FIXME: we should maybe log a warning

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+                return rule["authz"]
+
+        return None  #### FIXME this is just for testing purposes #####
```
Avantol13: let's remove this

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+        raise ValueError(
+            f"No authorization tag mapping found for resource ID: {resource.get('id')}"
+        )
+
+    def tag_resource(self, resource: dict, authz: str) -> dict:
+        """
+        Injects the authz path directly into the standard FHIR meta block
+
+
+        FIXME: match Gen3 FHIR Proxy service requirements here
```
Avantol13: instead of a FIXME, just say: "MUST match Gen3 FHIR..."

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+            resource (dict): tagged resource with the appropriate authz string
+        """
+
+        if "meta" not in resource:
+            resource["meta"] = {}
+        if "security" not in resource["meta"]:
+            resource["meta"]["security"] = []
+
+        gen3_security_tag = {
+            "system": "http://gen3.org/authz",
```
Avantol13: this path should be configuration or at least a global variable at the top of this file

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+        }
+
+        if gen3_security_tag not in resource["meta"]["security"]:
+            resource["meta"]["security"].append(gen3_security_tag)
+
+        return resource
+
+
+def json_dumps(obj, default=None) -> bytes:
+    """Compact UTF-8 JSON bytes. Stand-in for orjson.dumps."""
```
Avantol13:
```suggestion
    """Compact UTF-8 JSON bytes."""
```

no need to mention a package we ended up not using

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+    Returns:
+        status (bool): returns whether the directory is new/has to be reset or not
+    """
+
+    if not pathlib.Path(directory).is_dir():
+        return True
+
+    try:
+        params = json.loads(pathlib.Path(directory, ".config.json").read_bytes())
+    except:
```
Avantol13: don't blindly catch broad exceptions like this b/c this implicitly captures things like ctrl-c during command line execution. `except ExactExceptionYouExpect:`

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+        batch_size (int): batch size of each chunk
+        output_dir (str): the path to the directory where to write all the intermediate files to
+
+    """
+    if batch_size < 1:
+        raise ValueError(f"batch_size must be >= 1, got {batch_size}")
+    start = time.time()
+    with open(input_file, "rb") as fin:
+        for i, lines in enumerate(iter(lambda: list(islice(fin, batch_size)), [])):
+            # with open(f"chunk_{i:05d}.chunk", "wb") as out:
```
Avantol13: dead code, remove

gen3/fhir.py
```diff
@@ -0,0 +1,490 @@
...
+        os.path.join(output_dir, f"{os.path.basename(input_file)}.done"), "wb"
+    ) as fout:
+        out = bytearray()
+        for r in fin:
+
+            if not r.strip():
+                continue
+
+            record = json.loads(r)
+            authz_tags = tagger.determine_authz(record)  # generate tags
```
Avantol13: no in-line comments, only above the line

pyproject.toml
dsafarian: 1. Pytest version needs to be update to at least 7+
2. If I don't pin setuptools version, and it installs version 84, I get this error:
gen3sdk-python/tests/conftest.py'.
tests/conftest.py:11: in <module>
    from drsclient.client import DrsClient
../../../Library/Caches/pypoetry/virtualenvs/gen3-x8jGTXip-py3.13/lib/python3.13/site-packages/drsclient/__init__.py:1: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'

dsafarian: and in the original gen3sdk poetry lock there is this, so I think the setuptools verion has to be below 82 in the way it is written now:
[[package]]
name = "setuptools"
version = "80.9.0"
description = "Easily download, build, install, upgrade, and uninstall Python packages"
optional = false
python-versions = ">=3.9"
groups = ["main", "dev"]
files = [
    {file = "setuptools-80.9.0-py3-none-any.whl", hash = "sha256:062d34222ad13e0cc312a4c02d73f059e86a4acbfbdea8f8f76b28c99f306922"},
    {file = "setuptools-80.9.0.tar.gz", hash = "sha256:f36b47402ecde768dbfafc46e8e4207b4360c654f1f3bb84475f0a28628fb19c"},
]

gen3/cli/fhir.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Args:
+        input_file (str): Input .ndjson file
+        output_file (str): Output file name
+        config (str): .yaml file with authorization rules
+        batch_size (int): numb er of lines per chunk
+        workers (int): number of parallel processes
+        force (bool): remove all intermediate files for this run before exiting even if it crashes
+    """
+    fhir_tagger(input_file, output_file, config, workers, batch_size, force)
```
Avantol13: these are passed in the wrong order. I always recommend using explicit keywords in Python function calls for this reason. 

fhir_tagger(input_file=input_file, ... workers=workers, ...)

gen3/cli/__main__.py
```diff
@@ -15,6 +15,7 @@
 import gen3.cli.drs_pull as drs_pull
 import gen3.cli.users as users
 import gen3.cli.wrap as wrap
+import gen3.cli.fhir as fhir
```
Avantol13: this import needs to be optional b/c if someone does an install WITHOUT the `fhir` extra, this will fail. So you need to allow it missing here and then allow adding it later to fail and that's okay

pyproject.toml
Avantol13: try this file (it worked for me):

```yaml
[tool.poetry]
name = "gen3"
homepage = "https://gen3.org/"
version = "4.29.0"
description = "Gen3 CLI and Python SDK"
authors = ["Center for Translational Data Science at the University of Chicago <support@gen3.org>"]
license = "Apache-2.0"
packages = [
    { include = "gen3" },
]
classifiers = [
    "Development Status :: 4 - Beta",
    "Intended Audience :: Science/Research",
    "Intended Audience :: Developers",
    "License :: OSI Approved :: Apache Software License",
    "Programming Language :: Python :: 3.9",
    "Topic :: Scientific/Engineering",
]

[tool.poetry.dependencies]
python = ">=3.9, <4"
requests = "*"
indexclient = ">=2.3.0"
drsclient = ">=0.3.0"
aiohttp = "*"
backoff = "*"
cdislogging = ">=1.1.0"
click = "*"
importlib_metadata = ">=8,<9"
jsonschema = "*"
# FIXME updating to >=0.6.0 breaks a few tests
dataclasses-json = "<=0.5.9"
pypfb = ">=0.6.2"
tqdm = ">=4.61.2"
humanfriendly ="*"
python-dateutil = "*"
aiofiles = "*"
pandas = ">=1.4.2"
xmltodict = ">=0.13.0"
gen3users = "*"

# A list of all of the optional dependencies, some of which are included in the
# below `extras`. They can be opted into by apps.
fhirclient = { version = "*", optional = true }
fhirpathpy = { version = "*", optional = true }

[tool.poetry.extras]
fhir = ["fhirclient", "fhirpathpy"]

[tool.poetry.group.dev.dependencies]
pytest = ">=7.0"
pytest-cov = "*"
requests-mock = "*"
cdisutilstest = { git = "https://github.com/uc-cdis/cdisutils-test.git", tag = "2.0.0" }
indexd = { git = "https://github.com/uc-cdis/indexd.git", tag = "5.0.4" }
deptry = ">=0.23.1"

[tool.poetry.scripts]
gen3 = "gen3.cli.__main__:main"

[tool.poetry.plugins."gen3.plugins"]

[build-system]
requires = ["poetry>=1.4.1"]
build-backend = "poetry.masonry.api"

[tool.pytest.ini_options]
minversion = "6.0"
addopts = "-vv"
testpaths = [
    "tests",
]
```

tests/tests/outputs/empty_file
Avantol13, line 1: are all these tests/outputs necessary for your own testing or were they checked in by accident? These look unrelated and likely for other test outputs. Trying deleting them

Avantol13 (changes requested):

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,555 @@
...
+        for c in transformed
+    )
+    merge_chunks(transformed, OUT)
+    merged = pathlib.Path(OUT).read_text(encoding="utf-8")
+    fin = [
+        json.loads(line)
+        for line in pathlib.Path(IN).read_bytes().splitlines()
+        if line.strip()
+    ]
+    fout = fin = [
```
Avantol13: was this intentional? fin is now fout

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,555 @@
...
+def test_cli():
+    """Run the CLI and return the CompletedProcess."""
+    args = [
+        IN,
+        OUT,
+        CONFIG_SRC,
+        "--batch_size",
+        str(BATCH_SIZE),
+    ]
+    result = subprocess.run(
```
Avantol13, line 425: you should assert on a successful result here

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,555 @@
...
+    tag_fhir_resources_with_authz,
+    DEFAULT_WORK_DIR
+)
+import pathlib
+import os
+import json
+import subprocess
+import shutil
+
+TMP_ROOT = pathlib.Path("./tests/test_data/fhir_outputs")
```
Avantol13: we should look into https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html instead of a relative path here, b/c this requires you to run from root to get the outputs in the expected place to .gitignore

The side effect of this is exactly what I think were seeing with the tests/tests/outputs files you accidentally checked in. Those likely used relative paths from wherever the script was run, and you probably ran it with tests/ as the root instead of the repo root at some point.

You don't have to fix it there - but we shouldn't repeat that pattern here

gen3/fhir.py
```diff
@@ -0,0 +1,519 @@
...
+    root = pathlib.Path(
+        work_dir or os.environ.get("GEN3_FHIR_WORK_DIR") or DEFAULT_WORK_DIR
+    ).expanduser()
+   
+    if not clean:
+        root.mkdir(parents=True, exist_ok=True)
+
+    # 0o700 makes it only only readable by owner and not everyone else (which is important 
+    # for potentially shared machines and potential FHIR PHI)
+    root.chmod(0o700)
```
Avantol13: if root.is_dir():
   root.chmod(0o700)

we could maybe guard this to make sure it's a dir

gen3/fhir.py
```diff
@@ -0,0 +1,519 @@
...
+    return json.dumps(
+        obj,
+        ensure_ascii=False,
+        allow_nan=False,
+        separators=(",", ":"),
+        default=default,
+    ).encode("utf-8")
+
+
+def get_md5hash(input_file: str | os.PathLike[str]) -> str:
```
Avantol13: We should probably use sha256 - it's more secure and md5 is going to trip some security and compliance flags

gen3/cli/fhir.py
```diff
@@ -0,0 +1,127 @@
...
+@click.option(
+    "--work_dir",
+    type=click.Path(),
+    metavar="work_dir",
+    help=f"Specify which working directory to clean, if not specified the default ({DEFAULT_WORK_DIR}) will be cleaned. Can also be set as an environment variable: GEN3_FHIR_WORK_DIR",
+)
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    help="Used with --cleanup, report what would be deleted with --cleanup without deleting the files",
```
Avantol13: out of date. `--cleanup` -> `--force`

tests/test_data/fhir_config.yaml
```diff
@@ -0,0 +1,39 @@
+version: "1.0"
+
+
+# Global Override (Applies to all resources if used)
+# global_authz: "/programs/NCI/projects/TargetedStudyA"
```
Avantol13: we need some unit tests to cover global_authz

gen3/cli/__main__.py
dsafarian: But then this would also fail. How would i add the subcommand:

`main.add_command(fhir.fhir)`

Should I do an if statement / try&except for the import?

tests/test_fhir_transform.py
dsafarian: tests/tests/outputs -  these outputs are from the other tests, mine are linked to be within test_data/fhir_output, I forgot to delete the other ones. 

I did try to use the pytest tmp_path, the TMP_ROOT is just defining what i want that path to be. Maybe I'm still confused on how to use it:
`@pytest.fixture(scope="session")
def tmp_root(tmp_path_factory):
    return tmp_path_factory.mktemp(TMP_ROOT)`
    
 I then pass tmp_root to all the tests which require it

gen3/cli/__main__.py
Avantol13: Yes, we need it all behind conditionals b/c if you don't do that and someone installs it without the fhir extra, the imports will eventually fail b/c we don't include those deps. You can try this out in a fresh virtualenv. don't install the FHIR extra and try to use the CLI

tests/test_fhir_transform.py
Avantol13: can we try to do it without the root being specified?

`Temporary directories are by default created as sub-directories of the system temporary directory` the docs claim this

pyproject.toml
dsafarian: I checked again and this doesnt work...
```
ImportError while loading conftest 'gen3sdk-python/tests/conftest.py'.
tests/conftest.py:11: in <module>
    from drsclient.client import DrsClient
../../../Library/Caches/pypoetry/virtualenvs/gen3-x8jGTXip-py3.13/lib/python3.13/site-packages/drsclient/__init__.py:1: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
```

gen3/cli/__main__.py
```diff
@@ -146,5 +145,47 @@ def main(
...
+except ImportError:
+
+    @click.group(name="fhir")
+    def fhir():
+        """Commands for FHIR data processing: transform & cleanup"""
+        raise click.ClickException(
+            "Requires FHIR packages which aren't installed by default. Install the 'fhir' extras: poetry install --all-extras"
+        )
+
+    @fhir.command(
```
Avantol13: lets remove these and only have the root fhir command, that's enough for someone to understand it exists and what to do to install it. I'd rather not duplicate commands here

docs/howto/fhir.md
```diff
@@ -0,0 +1,33 @@
...
+`gen3 fhir COMMAND [ARGS] [OPTIONS]`
+
+For a list of commands and options run
+
+`gen3 fhir --help`
+
+For example, the following tags the 'Patient.ndjson' file with Gen3 authorization and outputs 'gen3_Patient.ndjson' using the authorization rules from 'config.yaml'
+
+`gen3 fhir transform Patient.ndjson gen3_Patient.ndjson config.yaml --batch_size 10000 --workers 8 `
```
Avantol13:
```suggestion
`gen3 fhir transform Patient.ndjson gen3_Patient.ndjson config.yaml --batch_size 10000`
```

docs/howto/fhir.md
```diff
@@ -0,0 +1,33 @@
+## FHIR 
+
+This integration aims to enhance the data ingestion capabilities of Gen3 by integrating a Fast Healthcare Interoperability Resources (FHIR) data ingestion pipeline & tools.  FHIR is an important standard for working with Electronic Health Records (EHR) and we have started development of a Gen3 FHIR Proxy service.  The overall goal is to allow users to seamlessly ingest data into an existing FHIR server. Gen3 is working on adding support for FHIR and this tool will help with data preparation and interaction in the future.
```
Avantol13:
```suggestion
This integration aims to enhance the data ingestion capabilities of Gen3 by integrating a Fast Healthcare Interoperability Resources (FHIR) data ingestion tools. FHIR is an important standard for working with Electronic Health Records (EHR) and we have started development of a Gen3 FHIR Proxy service. 

Gen3 is working on adding support for FHIR and these tools will help with data preparation.
```

docs/howto/fhir.md
```diff
@@ -0,0 +1,33 @@
...
+
+```yaml
+rules:
+ - resource_type: "Patient"
+   condition: "Patient.managingOrganization.reference = 'Organization/site-alpha'"
+   authz: "/programs/Alpha/projects/Main"
+
+#Example with multiple conditions
+ - resource_type: "Specimen"
+   condition: ""Specimen.status = 'available' and Specimen.Type = 'Blood specimen (specimen)'""
```
Avantol13: Is this supposed to have double double-quotes? "" ... "" ? looks like invalid YAML

docs/howto/fhir.md
```diff
@@ -0,0 +1,33 @@
...
+   authz: "/programs/Alpha/projects/Biobank"
+```
+
+And example config.yaml file can be found in [fhir_config.yaml](tests/test_data/fhir_config.yaml)
```
Avantol13: this link doesn't work from here, you need it to be `../../tests/test_data/fhir_config.yaml`

let's also add a complete example command using the test data:

```
poetry run gen3 -vv fhir transform ./tests/test_data/Patient.ndjson ./test_Patient.ndjson ./tests/test_data/fhir_config.yaml
```

gen3/cli/fhir.py
```diff
@@ -0,0 +1,122 @@
...
+@click.argument(
+    "config",
+    type=click.Path(exists=True, dir_okay=False, readable=True),
+    metavar="config",
+)
+@click.option(
+    "--work_dir",
+    type=click.Path(),
+    metavar="work_dir",
+    help=f"Specify which working directory to clean, if not specified the default ({DEFAULT_WORK_DIR}) will be cleaned. Can also be set as an environment variable: GEN3_FHIR_WORK_DIR",
```
Avantol13:
```suggestion
    help=f"Specify which working directory to clean, if not specified the default ({DEFAULT_WORK_DIR}) will be used. Can also be set as an environment variable: GEN3_FHIR_WORK_DIR",
```

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+        json.loads(line)
+        for line in pathlib.Path(IN).read_bytes().splitlines()
+        if line.strip()
+    ]
+    src = [
+        json.loads(line)
+        for line in pathlib.Path(SRC).read_bytes().splitlines()
+        if line.strip()
+    ]
+    tag_fhir_resources_with_authz(IN, OUT, CONFIG_SRC, BATCH_SIZE, TMP_ROOT, 8)
```
Avantol13, line 119: the CLI side of this is fixed, thanks. the same call in this test file is still positional though, and it's now actually wrong:

```python
tag_fhir_resources_with_authz(IN, OUT, CONFIG_SRC, BATCH_SIZE, TMP_ROOT, 8)
```

the signature is `(input_file, output_file, config, batch_size, work_dir, force)`, so that trailing `8` is binding to `force` - a leftover from `workers`. the test has been force-deleting the run dir every time it runs. this is the exact thing I was worried about 😅 let's make these keyword args:

```suggestion
    tag_fhir_resources_with_authz(
        input_file=IN,
        output_file=OUT,
        config=CONFIG_SRC,
        batch_size=BATCH_SIZE,
        work_dir=TMP_ROOT,
    )
```

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+SRC = pathlib.Path(
+    f"{pathlib.Path(__file__).parent}/test_data/test_fhir_Patient.ndjson"
+)
+IN = pathlib.Path(f"{pathlib.Path(__file__).parent}/test_data/Patient.ndjson")
+OUT = pathlib.Path(f"{TMP_ROOT}/fhir_output_Patient.ndjson")
+CONFIG_SRC = pathlib.Path(f"{pathlib.Path(__file__).parent}/test_data/fhir_config.yaml")
+GLOBAL_CONFIG_SRC = pathlib.Path(
+    f"{pathlib.Path(__file__).parent}/test_data/fhir_global_config.yaml"
+)
+TAGGER = Gen3FHIRAuthzTagger(CONFIG_SRC)
```
Avantol13, line 31: the makedirs is gone, thanks. the tagger is still constructed at import time though, and `Gen3FHIRAuthzTagger.__init__` opens and parses the config file - so importing this module touches the disk (which isn't great practice) and a collection error comes out of `gen3/fhir.py:49` is what you get if the path doesn't resolve.

use a fixture instead:

```python
@pytest.fixture(scope="session")
def tagger():
    """A tagger built from the synthetic Patient rules config."""
    return Gen3FHIRAuthzTagger(CONFIG_SRC)
```

gen3/cli/fhir.py
```diff
@@ -0,0 +1,122 @@
...
+)
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    help="Report what would be deleted with --cleanup without deleting the files",
```
Avantol13, line 101:
```suggestion
    help="Report what would be deleted without deleting the files",
```

--cleanup isn't here anymore

gen3/fhir.py
```diff
@@ -0,0 +1,605 @@
...
+        config_path: str | os.PathLike[str],
+        custom_hook: Callable[[dict], dict] | None = None,
+        fhir_version: str = "r4",
+    ):
+        """
+        Initialize instance of the tagger.
+
+        Args:
+            config_path (str): the name/path of the config.yaml file
+            custom_hook (Callable): used for authorization instead of the configuration if called
```
Avantol13, line 45: type hint's right, thanks. the docstring still doesn't say what the hook gets handed or what it's expected to give back, which is the part I was after - someone writing a hook can't tell from this.

also the annotation and the usage disagree: it's typed `Callable[[dict], dict]` but `determine_authz` returns whatever it produces as the authz, and that function is declared `-> str`.

```suggestion
            custom_hook (Callable[[dict], str] | None): if provided, called
                instead of the configuration to determine authz. Receives the
                FHIR resource as a dict and returns the authz path. A falsy
                return falls through to the configured rules.
```

docs/howto/fhir.md
```diff
@@ -0,0 +1,38 @@
...
+For a list of commands and options run
+
+`gen3 fhir --help`
+
+For example, the following tags the 'Patient.ndjson' file with Gen3 authorization and outputs 'gen3_Patient.ndjson' using the authorization rules from 'config.yaml'
+
+`gen3 fhir transform Patient.ndjson gen3_Patient.ndjson config.yaml --batch_size 10000`
+
+
+The authorization configuration file has to be in yaml format and can have multiple conditions, e.g:
```
Avantol13, line 21: better, thanks. still missing the things a person writing their own config would trip over:

- `global_authz` exists and overrides every rule - it's in the code and in tests/test_data/fhir_global_config.yaml but not on this page at all
- `condition` is FHIRPath. worth saying so and linking https://hl7.org/fhirpath/ b/c otherwise people will guess at the syntax
- two rules matching the same resource is now an error, not first-match-wins. that's a real constraint on how you write a config and it needs to be stated
- one resource type per file. it's in --help but not here

gen3/cli/__main__.py
```diff
@@ -145,4 +145,23 @@ def main(
...
+    main.add_command(fhir.fhir)
+except ImportError as e:
+
+    @click.group(
+        name="fhir",
+        epilog="Requires FHIR packages which aren't installed by default. Install the 'fhir' extras: poetry install --all-extras",
+    )
+    def fhir():
+        """Commands for FHIR data processing: transform & cleanup"""
+        raise
```
Avantol13, line 162: duplicates are gone, thanks.

the bare `raise` that replaced the ClickException doesn't work though. by the time the group callback runs there's no exception in flight, so it'd be `RuntimeError: No active exception to reraise` - and it never even gets there. I simulated the no-extras install:

```
gen3 fhir                    -> prints help (epilog does mention extras, good)
gen3 fhir --help             -> exit 0
gen3 fhir transform a b c    -> "No such command 'transform'."
```

so the one command we document is the one that gives a useless error. let's register a catch-all so any fhir subcommand says the right thing:

```python
except ImportError:

    @click.group(
        name="fhir",
        invoke_without_command=True,
        epilog="Requires FHIR packages which aren't installed by default. Install the 'fhir' extras: poetry install --all-extras",
    )
    @click.argument("args", nargs=-1, type=click.UNPROCESSED)
    def fhir(args):
        """Commands for FHIR data processing: transform & cleanup"""
        raise click.ClickException(
            "FHIR commands require the 'fhir' extras: poetry install --all-extras"
        )
```

(the `as e` on the except isn't used either so could be removed )

gen3/fhir.py
```diff
@@ -0,0 +1,605 @@
...
+        matches = []
+        matches = [
+            rule
+            for rule in self.rules
+            if (result := rule["match"](resource)) and result[0] is True
+        ]
+
+        if len(matches) > 1:
+            raise ValueError(
+                f"Resource {resource.get('id')!r} matched {len(matches)} authorization conditions; expected at most 1. Conflicting: {[m["condition"] for m in matches]}"
```
Avantol13, line 112: this is why 3.10 and 3.11 are failing. nesting the same quote style inside an f-string is 3.12+ only, and pyproject says we support >=3.10:

```suggestion
                f"Resource {resource.get('id')!r} matched {len(matches)} authorization conditions; expected at most 1. Conflicting: {[m['condition'] for m in matches]}"
```

same thing at tests/test_fhir_transform.py:179 and :211. worth knowing that our pinned black (23.1.0) can't parse these two files either - it errors out rather than formatting them, so the "Black formatting" commit skipped them silently. `poetry run black --check gen3/fhir.py tests/test_fhir_transform.py` shows it

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+    ), "The length of the output file does not math the sum of the .done files and the input file"
+    # check formatting of output file
+    assert merged.endswith("\n"), "File does not end with a newline"
+    assert not merged.endswith("\n\n"), "File ends with a blank line"
+    for i, line in enumerate(merged.split("\n")[:-1]):
+        assert line.strip(), f"blank line at index {i}"
+        json.loads(line)  # every line independently parses
+
+
+class Test_is_new:
```
Avantol13, line 224: this class and the other three never run. pytest won't collect a test class that has an `__init__`, and it only warns about it:

```
PytestCollectionWarning: cannot collect test class 'Test_is_new' because it has a __init__ constructor
```

`pytest --collect-only -q tests/test_fhir_transform.py` reports 12 tests. I dropped the four constructors in a scratch copy and 44 more showed up - so _is_new, _is_done, _merge_needed and the status-overlap tests have no coverage at all right now.

and one of them fails once it runs:

```
Test_is_new::test_is_not_new_when_config_matches
assert _is_new(directory, record) is False
AssertionError: assert True is False
```

that looks real rather than a fixture problem - mock_state(config="match") writes no output file, and _is_new now returns True when the output is missing or empty (gen3/fhir.py:274-278). so either the check is too aggressive or the test's expectation is stale, but it needs deciding.

use a class attribute or a fixture instead of the constructor:

```suggestion
class Test_is_new:
    """Tests the logic of the run/directory being marked as new (_is_new(directory, record) returns True)"""

    tmp_path = TMP_ROOT / "test_new"
```

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+        directory, record = mock_state(
+            self.tmp_path, config="match", chunks=chunk_files, done=done_files
+        )
+        assert _is_new(directory, record) is False
+        assert _is_done(directory, record) is False
+
+    @pytest.mark.parametrize(
+        "chunks", ["match", None, {"config_hash": "x"}, {"batch_size": 1}]
+    )
+    def test_merge_needed_overlap(self):
```
Avantol13, line 391: this parametrize doesn't match the function - it declares `chunks` and the test takes only `self`. right now it's masked by the class not being collected at all, but as soon as that's fixed it's a hard collection error and it takes all of Test_status with it:

```
ERROR In Test_status::test_merge_needed_overlap: function uses no argument 'chunks'
```

the same list is on test_new_and_done_are_mutually_exclusive as `state`, so I'm guessing this one is a copy-paste leftover and should just come off

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+    )
+    def test_new_and_done_are_mutually_exclusive(self, state):
+        # new and done are mutually exclusive
+        directory, record = mock_state(
+            TMP_ROOT / str(id(state)), config=state, output="full"
+        )
+        assert not (_is_new(directory, record) and _is_done(directory, record))
+
+
+def test_cli():
```
Avantol13, line 416: these tests shell out without --work_dir, so they run against DEFAULT_WORK_DIR and leave a run directory in the developer's real ~/.cache/gen3/fhir_transform, one per invocation, never cleaned up. we already have the flag - let's use a temp path and pass `--work_dir`.

gen3/fhir.py
```diff
@@ -0,0 +1,605 @@
...
+    """
+    working_dir = resolve_work_dir(work_dir, clean=True)
+    if not working_dir.is_dir():
+        logging.info("Nothing to clean")
+        return 0
+
+    # force to delete everything disregarding status
+    if force:
+        shutil.rmtree(working_dir, ignore_errors=True)
+        return
```
Avantol13, line 360: this returns None but the signature says `-> int`, so --force gives back nothing where --dry-run gives a count.

```suggestion
    if force:
        count = sum(1 for p in working_dir.glob("*") if p.is_dir())
        shutil.rmtree(working_dir, ignore_errors=True)
        return count
```

docstring's missing `work_dir` under Args too, and it says `Output:` where the rest of the file says `Returns:`

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+    assert {r["id"] for r in fin} == {
+        r["id"] for r in out
+    }, "IDs in the output file do not match the IDs in the input file. Order not maintained"
+    assert [{k: v for k, v in r.items() if k != "meta"} for r in out] == [
+        {k: v for k, v in r.items() if k != "meta"} for r in fin
+    ], "Content of the output file does not match the content of the input file (other than the security tags)"
+    assert src == out, "Output file does not match source file"
+
+
+def test_chunking():
```
Avantol13, line 139: these three (chunking, merge, transform) tests are a chain: test_chunking writes the .chunk files, test_transform turns them into .done, test_merge consumes those. so `pytest -k test_merge` on its own fails, and so does any parallel or randomized run - the inputs come from a sibling test rather than from the test itself. We can't have unit tests that depend on each other, we must make them independent or a giant combined test.

the chunk/done fixtures are cheap to build (mock_state already does most of it), so I'd rather each of the three set up what it needs rather than rely on each other. 

test_transform also calls relevant_authz_rules on the module-level TAGGER, which mutates it for everything that runs afterwards (we need to avoid that or ensure it gets set back to what it was before the test ran)

gen3/fhir.py
```diff
@@ -0,0 +1,605 @@
...
+            "display": f"Gen3 Policy Path: {authz}",
+        }
+
+        if gen3_security_tag not in resource["meta"]["security"]:
+            resource["meta"]["security"].append(gen3_security_tag)
+
+        return resource
+
+
+def resolve_work_dir(
```
Avantol13, line 153: a few spots in the new code are missing the docstrings and hints the rest of the module has. worth one pass over all of them:

- `resolve_work_dir` has no docstring, and it's public and does two surprising things - reads GEN3_FHIR_WORK_DIR, and chmods to 0o700. both belong in it
- `get_resource_type(input_file, sample=5)` and `json_dumps(obj, default=None)` have unannotated params
- `relevant_authz_rules` has no return annotation (`-> None`) where everything else in the class is annotated
- `transform_chunk` types the tagger as `object`; `Gen3FHIRAuthzTagger` is what it means
- in the tests, `mock_state` has no hints and its docstring documents a `tmp_path (Path)` arg - the parameter is actually `directory`

pyproject.toml
```diff
@@ -13,46 +13,46 @@ classifiers = [
     "Intended Audience :: Science/Research",
     "Intended Audience :: Developers",
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.9",
```
Avantol13, line 16: we need to also update .github/workflows/ci.yaml to remove 3.9 from the tests.

and we need to bump the major version now since we're dropping a whole Python version. so bump version above to `5.0.0` and add a Breaking Change in the PR description template explaining the removal due to it not being officially supported anymore and dependencies we need no longer support it

tests/test_fhir_transform.py
```diff
@@ -0,0 +1,551 @@
...
+BASE_RECORD = {
+    "timestamp": "2026-08-12T21:00:20.677168+00:00",
+    "input_file": f"{IN}",
+    "output_file": f"{pathlib.Path(TMP_ROOT)}/status.ndjson",
+    "config_hash": "9c61e70e3ea403e3daf1ee795036253c",
+    "batch_size": 1,
+}
+
+
+@pytest.fixture(scope="session")
```
Avantol13, line 42: this is great, but tests aren't using it, looks like they're still using TMP_ROOT directly. `scope="session"` only controls how long it's cached once something requests it, it doesn't make pytest call it. a fixture body runs when a test takes it as an argument, or when it's `autouse=True`

which matters b/c that fixture is the only thing that would wipe TMP_ROOT between runs. right now the directory just accumulates

gen3/fhir.py
```diff
@@ -0,0 +1,605 @@
...
+        or record["batch_size"] != params["batch_size"]
+        or record["output_file"] != params["output_file"]
+    ):
+        return True
+
+    if (
+        not pathlib.Path(record["output_file"]).is_file()
+        or pathlib.Path(record["output_file"]).stat().st_size == 0
+    ):
+        return True
```
Avantol13, line 278: sorry for the long comment - but this re-analyzes the overall resume functionality in depth and has some broader suggestions.

firstly: the resume path here can't ever fire, and since the plan is to build parallelization on top of this eventually I'd rather we fix the foundation in this PR.

`_is_new` returns True when the output file is missing or empty here. but that's the normal state of an interrupted run, so it fights `_merge_needed`, and the pipeline checks `_is_new` first (line 494) and rmtree's the directory. I set up a crashed-run state - matching .config.json, three .done files, no output yet - and ran the pipeline:

```
_is_done      -> False
_merge_needed -> True    <- "resume: just merge"
_is_new       -> True    <- "throw it away and restart"
[INFO] New folder, creating directory and chunking
pre-existing .done files that survived: NONE
```

so it's either "already done, skip" or "start from scratch" - the resume branch is dead. output is still correct, it just always redoes the whole file. and `test_is_not_new_when_config_matches` is the hidden test that actually fails once you make the class collectible: it was telling us this.

four things I'd do, roughly in order of value:

**1. give `_is_new` one job.** it's answering two questions - "are this directory's parameters still valid" and "is the final output already there". only the first belongs to it; the second is `_is_done`'s. then order the checks so resume wins:

```python
if _is_done(output_dir, record):      # output exists and params match -> nothing to do
    return
if not _params_match(output_dir, record):   # config/batch_size/output path/input changed
    reset_run_dir(output_dir, record)
# otherwise: pick up whatever units are already .done and finish the rest
```

**2. make "done" atomic.** `transform_chunk` (line 417) opens `<chunk>.done` and writes at the very end, so a crash between the open and the write leaves a zero-length .done that later looks complete. write `<chunk>.done.tmp` and `os.replace()` it into place - atomic within a filesystem, so a .done that exists is always whole. right now that's an unlikely race; with W workers it becomes routine, and every resume decision is built on trusting these files.

**3. don't hold a unit in memory if we don't need to.** `transform_chunk` accumulates the whole chunk in a bytearray before one `fout.write`. writing per record should let the buffered file object do the batching itself

**4. consider dropping the chunk *files* while keeping the chunk *abstraction*.** a work unit doesn't have to be a file - it can be a `(start_offset, end_offset)` byte range over the input, split on newline boundaries, with each worker reading its own slice. that kills the serial `split_file` pre-pass (today nothing gets tagged until the entire input has been copied to scratch), takes peak scratch from ~2x input down to 1x, and makes unit boundaries a recomputable function of the input instead of on-disk state that can be half-written. this is the biggest of the four and the least urgent - 1 through 3 matter more.

also: docstrings say "Stream-transform" but this isn't truly streaming so I'd remove that.

Avantol13 (changes requested):
see comments

tests/test_fhir_transform.py
dsafarian: This goes back to what we had discussed in a previous meeting when we decided we didn't want the directory to be wiped all the time since the files would be useful for the tests?

Avantol13: so a couple things:

- if that's the case, we should just remove this fixture b/c it's unused then
- however, I think it might actually be better to walk that back and change our minds. By default we should clear everything out by using this fixture and if someone needs to inspect files in the middle of a test, they can add temporary code to avoid deleting (e.g. comment it out in the fixture)

PR #314, opened 2026-08-20
Feat/dpop

Avantol13:
Depends on:
- https://github.com/uc-cdis/authutils/pull/113
- https://github.com/uc-cdis/fence/pull/1372
⚠️ Merge those first and then update authutils here! 

### New Features
- DPoP support (CLI & SDK)
- Nextflow run support utilizing DPoP (CLI & SDK)

### Breaking Changes
- 3.13 only due to required updates from upstream dependencies, including authutils and other Gen3 libraries (this matches the Python version used and required by other Gen3 services) 

### Improvements
- Unit tests parallelized, down to ~7s from ~40s locally (w/ the new tests included) - looks like existing runs would take ~30s before these new tests and on this branch they take ~17s, so still almost half the time with many new tests added
- Rip out coupling to indexd by actually mocking in unit tests


docs/howto/nextflow.md
```diff
@@ -0,0 +1,276 @@
...
+workDir = 's3://<your-workflow-bucket>/<your-prefix>'
+```
+
+Run that with `gen3 nextflow run main.nf` and the generated override changes five keys:
+`tes.endpoint` and `aws.client.endpoint` become the local proxy, `tes.oauthToken` and
+`aws.accessKey` become the task token, and `aws.client.s3PathStyleAccess` is forced on.
+`workDir`, `region`, `timeout`, `maxErrorRetry`, `plugins` and `process` are yours and are
+left as they are. `gen3 -vv nextflow run` logs the generated file's contents; `nextflow -c
+
+<that file> config` prints the merged result.
```
paulineribeyre, line 95:
```suggestion
left as they are. `gen3 -vv nextflow run` logs the generated file's contents; `nextflow -c <that file> config` prints the merged result.
```

gen3/dpop.py
```diff
@@ -0,0 +1,996 @@
...
+                ASGI scope.
+            service (Service): Which upstream this is going to, from
+                `_resolve_upstream_url`.
+
+        Returns:
+            dict[str, str]: Headers to send upstream.
+        """
+        is_s3 = service == _SERVICE_S3
+        not_forwarded = (
+            _HEADERS_NOT_FORWARDED_TO_S3 if is_s3 else _HEADERS_NOT_FORWARDED
```
paulineribeyre, line 426: This is a bit redundant since _HEADERS_NOT_FORWARDED_TO_S3 is just _HEADERS_NOT_FORWARDED except "authorization", and the "authorization" header is overwritten anyway a couple lines later if "not is_s3".

Could we just remove "authorization" from _HEADERS_NOT_FORWARDED and get rid of _HEADERS_NOT_FORWARDED_TO_S3?

gen3/dpop.py
```diff
@@ -0,0 +1,996 @@
...
+            base = self._config["TES_ENDPOINT"].rstrip("/")
+            subpath = path.removeprefix("/ga4gh/tes")
+        elif path.startswith("/s3"):
+            service = _SERVICE_S3
+            base = self._config["S3_ENDPOINT"].rstrip("/")
+            subpath = path.removeprefix("/s3")
+        else:
+            logging.warning(
+                f"Refusing to proxy {path}: only /ga4gh/tes and /s3 paths are "
+                "proxied. Check the endpoints your pipeline is configured with."
```
paulineribeyre, line 380: The paths start with either `/ga4gh/tes/v1` or `/workflows`, see config [here](https://github.com/uc-cdis/gen3-helm/blob/9a33ce8/helm/revproxy/gen3.nginx.conf/gen3-workflow-service.conf).

Plus, the S3 endpoint is also exposed at the app root (see [here](https://github.com/uc-cdis/gen3-workflow/blob/f9a4c31/gen3workflow/routes/s3.py#L199-L200)).

So a request to Gen3 S3 could start with `/workflows` or `/workflows/s3` or `/ga4gh/tes/v1` or `/ga4gh/tes/v1/s3` (although the last 2 are not documented or used). But never just `/s3`.

I'm not sure what a reliable way to identify S3 requests while maintaining root S3 endpoint support would be. We could list [all the non-S3 routes](https://github.com/uc-cdis/gen3-workflow/blob/18fe9d6/gen3workflow/app.py#L77-L81) but that's not very future-proof.

For now to unblock my testing, i made [this change](https://github.com/uc-cdis/gen3sdk-python/commit/44b0523353eb90834954b0fd9c33abb4e12207fe), which assumes S3 requests are non-root.

Edit: I see the proxy is accepting `/s3` requests and forwarding them to `/workflows/s3`, and `DPOP_PROTECTED_PATHS` in gen3-workflow matches that, so maybe I misunderstood the intent. We can discuss it when you're back!

gen3/dpop.py
Avantol13: yeah, good call. I cleaned it up and added a check b/c of case-sensitivity causing duplicates

gen3/dpop.py
Avantol13: the proxy itself has its own endpoints and then translates those to the real Gen3 Workflow, and I'd rather keep them separate and have the mapping in the proxy itself. We can control in the Nextflow config what _local proxy_ endpoint to hit for S3 and TES respectively (and the auto-generated config should already be doing that). e.g. this should've worked out of the box

in your edit: `/ga4gh/tes` and `/s3` are the proxy's own local namespace, not commons paths. the commons paths only appear in `TES_ENDPOINT` / `S3_ENDPOINT`, and the generated nextflow config points nextflow at the _local_ ones.

round trip for S3: nextflow hits `127.0.0.1:port/s3/bucket/key`, proxy strips `/s3` and appends to `{commons}/workflows/s3`, signs htu over `{commons}/workflows/s3/bucket/key`.

gen3-workflow should rebuild that same string to check.

**What was the config that sent `/workflows/...` at the proxy? e.g. why didn't what was written work out of the box?**

if it was a hand-written pipeline config or the auto-generated config... we could fix that instead.

re: your changes, if we want to keep this and support commons paths matching locally - I think it needs some updates either way. It looks like this will happen:

```
/workflows/s3/bucket/key.txt  -> ('https://cx/workflows/s3/bucket/key.txt', 's3')
/workflows/bucket/key.txt     -> ('https://cx/ga4gh/tes/bucket/key.txt', 'tes')
/s3/bucket/key.txt            -> None
```

the middle one is the root-mounted S3 case you raised - after `/workflows/` matches, anything that isn't `/s3/` falls through to the TES base, so it goes to
`{commons}/ga4gh/tes/<bucket>/<key>` and since the service isn't s3 we replace the
SigV4 header with `Authorization: DPoP <token>`. and the last one is what the
generated config sends, so generated-config runs 404.

But... I still am not convinced we need to change this. The local proxy and nextflow config can be opinionated - we can choose to only support /s3 -> /workflows/s3 instead of all 4 options the service itself supports.

paulineribeyre: My integration tests do not use the path that updates the config automatically. I see that I misconfigured `aws.client.endpoint` (`<proxy>/workflows/s3` instead of `<proxy>/s3`), that's probably where the issue came from. I'll fix that and test - no need to change anything if that works 👍 

Maybe a bit of documentation/docstring about this endpoint mapping would be nice though, so the next reader isn't confused like I was?

paulineribeyre: Ok that was only part of the issue - all the existing TES tests were also pointing at `/workflows/s3`, which i didn't change when i updated them to use the dpop proxy.

PR #315, opened 2026-08-24
Testing external contributor PR #313

PlanXCyborg:
*** DO NOT MERGE ***
This PR was created automatically to test an external PR
Tested PR link [here](https://github.com/uc-cdis/gen3sdk-python/pull/313)


gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+
+        elapsed_time = time.time() - start_time
+        logging.info(f"Tagged file -> {output_file}")
+        logging.info(
+            f"Total time to process: {(time.strftime('%H:%M:%S', time.gmtime(elapsed_time)))}"
+        )
+        if force:
+            shutil.rmtree(output_dir, ignore_errors=True)
+
+    except BaseException as e:
```
Avantol13: You can't use BaseException here, it will completely absorb a CTRL+C from the terminal (KeyboardInterrupt). Ideally you shouldn't wrap so much code in a broad exception block. I would break the code out into a helper function so it's a bit cleaner to read, and while we don't encourage broad exception catching, if you do it, you should use the Exception class instead of BaseException.

Avantol13: also this needs to exit(1) after all the logging. This currently shows a "success" exit message

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+                shutil.copyfileobj(fin, fout)
+    logging.info(
+        f"merge chunks total time: {(time.strftime('%H:%M:%S', time.gmtime(time.time() - start)))}"
+    )
+    # remove .done files once merge completed
+    for f in input_files:
+        os.remove(f)
+
+
+def fhir_tagger(
```
Avantol13: we should rename this function to describe better the verb / action it takes. Currently it sounds like a class name (a thing that does something). I would argue for something like `tag_fhir_resources_with_authz`

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+                json.dumps(record, default=str), encoding="utf-8"
+            )
+            # split into chunks
+            logging.info(f"Chunking {input_file} into {batch_size}-sized batches...")
+            split_file(input_file, batch_size, output_dir)
+
+        # initialize tagger
+        tagger = Gen3FHIRAuthzTagger(config_path=config)
+        # keep only relevant rules for the resource type
+        tagger.relevant_authz_rules(os.path.basename(input_file).split(".")[0])
```
Avantol13: This means that we rely explicitly on the filename for the FHIR resource type instead of checking the actual `resourceType` in the resource itself... I think we should reconsider this and instead read the first few lines of the file (and maybe last few) - you'll need this to be done without reading the whole file for efficiency, then check for that `resourceType` of those lines and if all the same: use that. and then make sure it's clear in docs that we expect 1 resource type per file

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+
+        """
+
+        # use custom hook if provided
+        if self.custom_hook:
+            hook_result = self.custom_hook(resource)
+            if hook_result:
+                return hook_result
+
+        # check global catch-all fallback
```
Avantol13: it's not really a fallback b/c the rules never evaluate. it's an override

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+    Output:
+        count(int): number of directories deleted
+    """
+
+    if not TMP_ROOT.is_dir():
+        logging.info("Nothing to clean")
+        return 0
+
+    # force to delete everything disregarding status
+    if force:
```
Avantol13, line 358: we should maybe only do this if --dry-run was not passed? I don't think it make sense to delete everything if it's a dry run - but if there's an argument for it, I'm open to it

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+
+            record = json.loads(r)
+            # generate tags
+            authz_tags = tagger.determine_authz(record)
+            # tag resources
+            out += json_dumps(tagger.tag_resource(record, authz_tags))
+            out += b"\n"
+        fout.write(out)
+        # delete chunk file once it has been transformed
+        os.remove(input_file)
```
Avantol13: ooh, don't do this inside the with block of the file opening, there's intentional cleanup that happens after the with block is done. You should be able to just do this outside the with block

docs/howto/fhir.md
```diff
@@ -0,0 +1,30 @@
...
+
+```yaml
+rules:
+ - resource_type: "Patient"
+   condition: "Patient.managingOrganization.reference = 'Organization/site-alpha'"
+   authz: "/programs/Alpha/projects/Main"
+
+ - resource_type: "Patient"
+   condition: "Patient.managingOrganization.reference = 'Organization/site-alpha' and Patient.gender = 'male'"
+   authz: "/programs/Alpha/projects/Main"
```
Avantol13: I think this example fails to load b/c of Patient matching 2 authorization conditions, let's have spearate blocks to show each and have a single, good config someone could copy and use right away. And actually an example file or link to ones used in tests would be good too

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+
+    Stream-transform Bulk FHIR data to Gen3 compatible data with authorization tagging.
+    Parallelized line by line tagging with batched I/O
+
+    Args:
+            input_file (str): Input .ndjson file
+            output_file (str): Output file name
+            config (str): .yaml file with authorization rules
+            batch_size (int): number of lines per chunk
+            workers (int): number of parallel processes
```
Avantol13: this is not accurate - this gets passed through and used for multi-_threading_, not processing, and we don't get a benefit from multi-threading here b/c we're CPU-bound. Let's just remove the workers entirely and we can add multi-processing later in a separate PR to simplify this a bit.

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+        and pathlib.Path(record["output_file"]).is_file()
+        and pathlib.Path(record["output_file"]).stat().st_size != 0
+    ):
+        logging.info(f"File already tagged. Locate file here: {record['output_file']}")
+        return True
+
+    return False
+
+
+def _resume_run(directory: str | os.PathLike[str], record: dict) -> bool:
```
Avantol13: this is never called by real code, we need to remove it and fix the tests to not rely on it

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+from cdislogging import get_logger
+from gen3.utils import make_folders_for_filename
+from collections.abc import Callable
+
+logging = get_logger(__name__)
+TMP_ROOT = pathlib.Path(".fhir_transform/tmp")
+GEN3_PATH = "http://gen3.org/authz"
+
+
+class Gen3FHIRAuthzTagger:
```
Avantol13, line 22: a couple edge cases I think we need to handle:

1.  Delete or move output_file and re-run with the run dir intact: _is_done is False (b/c output gone) but _is_new is also False (.config.json matches), so nothing re-chunks. _merge_needed is False, so the transform block runs against an empty *.chunk glob and does nothing. merge_chunks([], output_file) then opens the output "wb", which is empty. Instead, it should recreate the missing file OR error with something like "already completed but output file missing. Run again with --force to create"

2. The _is_new reset path writes a fresh .config.json and re-splits but never clears old .chunk/.done files, so stale files survive a config change. You should clear out the run dir

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+        Args:
+            config_path (str): the name/path of the config.yaml file
+            custom_hook (Callable): used for authorization instead of the configuration if called
+        """
+
+        with open(config_path, "r") as f:
+            self.config = yaml.safe_load(f)
+        self.custom_hook = custom_hook
+
+    def relevant_authz_rules(self, resource_type: str):
```
Avantol13, line 54: We can make this even faster b/c fhirpathpy allows you to "compile" rules once and apply them over and over (which saves a good chunk of time). https://github.com/beda-software/fhirpath-py#compile

`evaluate(resource, "Patient.gender = 'male'")` takes the expression as a string and re-parses it every call. compile() does the parse once and hands back a callable that just applies the pre-parsed rule. Right now we reparse on every record for every rule.

You could build the callables once in relevant_authz_rules (where the rules are already being filtered) and call them in determine_authz instead of `evaluate(resource, rule["condition"])`.

gen3/cli/fhir.py
```diff
@@ -0,0 +1,106 @@
...
+
+@click.group()
+def fhir():
+    """Commands for FHIR data processing"""
+    pass
+
+
+@click.command(
+    context_settings={"help_option_names": ["-h", "--help"]},
+    help="Tag Bulk FHIR data with Gen3 compatible authorization tags",
```
Avantol13: You should add more detail here about the input_file, what type - what should be in it, etc
samre for output_file and config. The user needs all the detail about what to provide when they run ` poetry run gen3 fhir transform --help`

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+        raise ValueError(f"workers must be >= 1, got {workers}, default is 8")
+    # check if input and output file are the same
+    if os.path.realpath(input_file) == os.path.realpath(output_file):
+        raise click.UsageError("input_file and output_file must be different")
+
+    # only necessary the first time its run
+    make_folders_for_filename(TMP_ROOT)
+
+    # make temp directories
+    hash = get_md5hash(input_file)
```
Avantol13: we may want to think through how to check _is_done before hashing where possible. B/c if this input is like 100GB, this is going to take a while even if it's already been run previuosly. Maybe that's okay

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+    if batch_size < 1:
+        raise ValueError(f"batch_size must be >= 1, got {batch_size}")
+    if workers < 1:
+        raise ValueError(f"workers must be >= 1, got {workers}, default is 8")
+    # check if input and output file are the same
+    if os.path.realpath(input_file) == os.path.realpath(output_file):
+        raise click.UsageError("input_file and output_file must be different")
+
+    # only necessary the first time its run
+    make_folders_for_filename(TMP_ROOT)
```
Avantol13: don't do this b/c I think this treats the final /tmp as a filename. 

just do: `TMP_ROOT.mkdir(parents=True, exist_ok=True)`

gen3/fhir.py
```diff
@@ -0,0 +1,499 @@
...
+import shutil
+import pathlib
+import hashlib
+from concurrent.futures import ThreadPoolExecutor
+from cdislogging import get_logger
+from gen3.utils import make_folders_for_filename
+from collections.abc import Callable
+
+logging = get_logger(__name__)
+TMP_ROOT = pathlib.Path(".fhir_transform/tmp")
```
Avantol13: We shouldn't define this twice (here and CLI). ALso note that this is going to be scoped to the current directory
So resume only works if you re-run from the same directory, every directory you ever run from accumulates state, and cleanup only really sees the current directory.

Let's change it to a static location and allow it to be configured by an env var or passed argument (--work-dir). And maybe make a helper to get the right working dir.

```
DEFAULT_WORK_DIR = "~/.cache/gen3/fhir_transform"

def resolve_work_dir(work_dir: str | os.PathLike[str] | None = None) -> pathlib.Path:
    root = pathlib.Path(
        work_dir or os.environ.get("GEN3_FHIR_WORK_DIR") or DEFAULT_WORK_DIR
    ).expanduser()

    root.mkdir(parents=True, exist_ok=True)

    # 0o700 makes it only only readable by owner and not everyone else (which is important 
    # for potentially shared machines and potential FHIR PHI)
    root.chmod(0o700)

    return root
```

We'll need to update fhir cleanup (and everywhere else that was relying on this path) to go through this same resolver

Avantol13 (changes requested):

Avantol13 (changes requested):
see comments

gen3/fhir.py
dsafarian: The force option is for someone to pass in a normal run if they don't want any intermediates remaining. Otherwise, all intermediates remain in the temp folder. The dry-run option is in cleanup, where the user can choose whether they just want a list of files that would be deleted if they were to run the cleanup. Let me know if you want the dry-run as an option for the main pipeline as well.

gen3/fhir.py
dsafarian: I changed the ProcessPoolExecutor because running it within gen3 was causing errors. So it went from this:

`with ProcessPoolExecutor(max_workers=workers) as pool:`

to this:

`with ThreadPoolExecutor(max_workers=workers) as pool:`

If I understand correctly, it might be because the main() is a bare call within the gen3 cli. Let me know what I should do about this. 

This is the error:
`[2026-08-24 16:47:42,032][  ERROR] 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

        To fix this issue, refer to the "Safe importing of main module"
        section in https://docs.python.org/3/library/multiprocessing.html
        
[2026-08-24 16:47:42,032][  ERROR] run failed; intermediates left in .fhir_transform/tmp/Patient_4460d7077b0c90cdc50452d577ab8f6d
[2026-08-24 16:47:42,033][  ERROR] 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

        To fix this issue, refer to the "Safe importing of main module"
        section in https://docs.python.org/3/library/multiprocessing.html
        
[2026-08-24 16:47:42,033][  ERROR] run failed; intermediates left in .fhir_transform/tmp/Patient_4460d7077b0c90cdc50452d577ab8f6d
[2026-08-24 16:47:42,300][  ERROR] A process in the process pool was terminated abruptly while the future was running or pending.
[2026-08-24 16:47:42,301][  ERROR] run failed; intermediates left in .fhir_transform/tmp/Patient_4460d7077b0c90cdc50452d577ab8f6d`

gen3/fhir.py
Avantol13: nope it's okay to leave it out

gen3/fhir.py
Avantol13: let's just make this synchronous for now, or at a minimum correct the docs / code / comments to make it clear that this is multi-threading and generally CPU-bound so a ton more threads doesn't necessarily mean faster output

dsafarian: I commented out anything to do with parallelization for now and made it asynchronous in case we figure out a way to parrallelize. If not, I'll completely remove it.

gen3/fhir.py
dsafarian: What if for 1. one of the conditions for _is_new is if the output file isnt found?

gen3/fhir.py
dsafarian: In the case that someone hits CTRL+C at the start of the run, do we want to treat it as a force delete and delete the posisbly corrupt files immadiately?


The reason for the long try/except block and the broad exception is to be able to remove files if --force is passed at any point in the run for whatever the reason it is that the run failed

gen3/fhir.py
Avantol13: I think that would work 🤔 if it isn't found, then it's new, so clear out the dir and rerun everything

gen3/fhir.py
dsafarian: Done, but we can also check the resource type per record in transform chunk if necessary since we are already loading it

gen3/fhir.py
Avantol13: no, ctrl+c is a kill immediately request from the user, so we should avoid capturing and doing anything else - consequences are the user's. So if they do that, then they may have to run force the next time for a clean run. 

The broad exception is maybe okay, I'd just break it out to a helper function

